PR #27, opened 2017-04-27
Remove ng version

### LINKED ISSUE #26: Remove `ng-version` from snapshots
See: https://github.com/thymikee/jest-preset-angular/issues/23#issuecomment-296428230

thymikee:
**Summary**

Fixes #26.
Also, prepares for a bugfix in how snapshot serializers are invoked in Jest 20. 
As for now snapshots won't show Text nodes.

### DIFF
--- a/HTMLElementPlugin.js
+++ b/HTMLElementPlugin.js
@@ -13,7 +13,7 @@ function escapeHTML(str) {
   return str.replace(/</g, '&lt;').replace(/>/g, '&gt;');
 }
 
-const HTML_ELEMENT_REGEXP = /(HTML\w*?Element)|Text|Comment/;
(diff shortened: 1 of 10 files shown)


samgregory88:
@thymikee It's happening again in v4 looks like

thymikee:
That's weird, because the [snapshot](https://github.com/thymikee/jest-preset-angular/blob/master/example/src/app/calc/__snapshots__/calc.component.spec.ts.snap) doesn't reflect that, and the tests are passing.

thymikee:
Are you using Jest v21? The example is running on v21.0.1 currently.

samgregory88:
Ok i see the difference, i was using `fixture.debugElement.nativeElement` instead of just fixture. Is this expected if you use the nativeElement?

Issue #41, opened 2017-06-02
Webpack require() assets


cgatian:
Looks like this issue has been answered several times on the internet but nothing seems to work for me.

The approach I tried was to redirect the require to a mock, as suggested by the [Jest Webpack documentation](http://facebook.github.io/jest/docs/en/webpack.html#handling-static-assets)
```
"moduleNameMapper": {
      "\\.(jpg|jpeg|png)$": "<rootDir>/__mocks__/fileMock.js"
    },
```
This approach seems to work for the majority of users, but for some reason isn't for me.  I understand all setups are a little different, but I'm wondering if anyone has ran into this issue while running the preset, and how they solved it.

thymikee:
Hard to tell without a repo where the problem happens. Are you able to provide such? Otherwise there's little I can do

cgatian:
In your Angular CLI project example if you attempt to `require` a png file within a component that's under test, you should see the issue. 

I will attempt to fork your repo to show later tonight.

thymikee:
Can you send a PR with failing test which we can fix later?

PR #42, opened 2017-06-09
issue 41 example

cgatian:
This change isn't effected directly by the preset. However, it's an issue others my encounter when trying to use Webpack's require to import assets.

#41


thymikee:
Hey @cgatian, thanks a lot for this repro 👍 . It helped me to track a bug, which will be fixed in next major version, as it's a breaking change.
Closing in favour of #53

PR #49, opened 2017-06-13
Add typings for toMatchSnapshot

charpeni:
image


thymikee:
Hi @charpeni. Thanks for your contribution, but it's already available in `@types/jest`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jest/index.d.ts#L272

charpeni:
Oh thanks you're right. 

`@types/jasmine` was overlapping with `@types/jest`, removing types for jasmine fix the issue.

PR #71, opened 2017-09-14
be a bit more specific with the moduleNameMapper regexps

zerdos:
If you use the jest-preset-angular for an ionic project, there they have an app.ts, and the moduleNameMapper maps it to an incorrect location.


thymikee:
Wanted to fix this some time ago but forgot about it. Thank you!

zerdos:
yea - just tried it and now it works :)

PR #78, opened 2017-10-05
add default html to test environment

patrickmichalina:
In order to fix Angular Material tests, a default HTML shell should be used

This change fixes the following console warnings

```shell
 PASS  src/client/app/+admin/admin.component.spec.ts
  ● Console

    console.warn node_modules/@angular/material/bundles/material.umd.js:206
      Current document does not have a doctype. This may cause some Angular Material components not to behave as expected.
    console.warn node_modules/@angular/material/bundles/material.umd.js:206
      Current document does not have a doctype. This may cause some Angular Material components not to behave as expected.

```

screen shot 2017-10-05 at 12 33 40 pm

### DIFF
--- a/testEnvironment.js
+++ b/testEnvironment.js
@@ -12,10 +12,8 @@ class JSDOMEnvironment {
   constructor(config) {
     // lazy require
     const {JSDOM} = require('jsdom');
-
-    const html = '<!DOCTYPE HTML><html lang="en"><head><title>Test Environment</title></head><body></body></html>'
     
-    this.document = new JSDOM(html, {
+    this.document = new JSDOM('<!doctype html>', {
       url: config.testURL,
       runScripts: 'dangerously'
     });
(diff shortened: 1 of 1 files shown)


testEnvironment.js
```diff
@@ -13,7 +13,9 @@ class JSDOMEnvironment {
     // lazy require
     const {JSDOM} = require('jsdom');
 
-    this.document = new JSDOM(undefined, {
+    const html = '<!DOCTYPE HTML><html lang="en"><head><title>Test Environment</title></head><body></body></html>'
```
thymikee: Oh, forgot about that! I think we only need the doctype, no?

patrickmichalina: I think that would suffice! Maybe remove the <title></title>

thymikee: `new JSDOM('<!doctype html>, {` should be good :)

patrickmichalina:
Ok, I updated the branch

thymikee:
4.0.1 published, enjoy!

PR #84, opened 2017-10-18
Update README to include installation instructions.

brianmcallister:
Noticed this was missing from the README, although it is in the [blog post](https://www.xfive.co/blog/testing-angular-faster-jest/) . I thought it would make sense to include this here too as well.


thymikee:
Makes sense, thank you!

Issue #86, opened 2017-10-26
snapshot tests have a lot of whitespace


Haroenv:
Some trimming might happen here: https://github.com/thymikee/jest-preset-angular/blob/master/AngularSnapshotSerializer.js, not too sure where to start

thymikee:
The whitespace comes mostly from Jest's HTMLSerializer not clearing it up. It used to, but after Mark's adjustments it was gone. There was probably some good reason for it, but I don't remember. 
cc @pedrottimark

pedrottimark:
Yes, in https://github.com/facebook/jest/pull/4275 the plugin stopped trimming because it can remove content:
* content **space** at the beginning of a text node
* content **newline** (for example, in **pre** or **textarea** elements)
* an **empty** text node (which I think was possible in earlier versions of React)

Now it does look messy if content has line breaks and indentation like this test file:
https://github.com/SimenB/pretty-format-crash/blob/master/something.html

Y’all can fill a gap in my knowledge: does Angular render “insignificant” space to DOM?

thymikee:
@pedrottimark looks like this, that's why I initially went with just trimming it.

ewantavenerdemo:
Hi, what is the situation with this? I have recently implemented this package for my angular application but i have noticed a lot of whitespace in between elements. In some cases there are 4 blank lines in between elements. I can see it's coming from the AngularSnapshotSerializer.js.

Maybe there is something I can do to fix this.

pedrottimark:
@ewantavenerdemo Because I came to https://github.com/facebook/jest/tree/master/packages/pretty-format package from experience with React instead of Angular, some help is welcome.

When React renders to browser DOM or test environment, it does not add any extra space or newline between elements. Therefore the `DOMElement` plugin does not remove space or newline at the beginning of a line, because it is significant.

Does Angular render “pretty printing” newline and indentation space:
* to browser DOM? always? by default, but is there an option to omit?
* in test environment only? always? by default, but is there an option to omit?

ewantavenerdemo:
Ok. I've been having a look at the AngularSnapshotSerializer.js and seeing if i can resolve the formatting issue.

The original content before 'print' is called has lots of '/n'. When you call print (is this pretty-format?) you need to pass some configurations. I've tried to do this locally however when I set the 'min: true' in the config it doesn't seem to do anything. If I do a replace outside it will remove the '/n's but your own formatting by the serializer doesn't work.

[...]

PR #97, opened 2017-11-28
Adding serializer to remove internal angular attributes from the snapshots

sharikovvladislav:
Hi!

I added serializer to remove the internal angular attributes from the result snapshots.

This is good because snapshots become more human readable. This is also good because sometimes angular renderer changes the indexes of that internal attributes.

The questions:
1) This change is HUGE BC change. So a lot of tests for components which are not simple as one line will be broken. do you mind bumping the major version of jest-preset-angular? It is also possible to add some flag which will enable this logic. So this change will not influence existing users.
2) I see you have CI. I think it should be improved a bit. Do you mind triggering your build on releasing new versions of Angular? If you do so, the time of understanding that something went wrong will be decreased significantly. What can go wrong? For example, Angular released a new version where they added some new property which is not covered by my serializer. It must be added. Also, this should be mentioned in the docs, I think...


sharikovvladislav:
This is absolutely not final changes. It is just PR for showing the general idea of what I suggest. Let's say I created this for some discussion.

Technical stuff is really easy so we will fix it in a few moments.

example/src/app/a-bit-more-complex-component/__snapshots__/a-bit-more-complex.component.spec.ts.snap
```diff
@@ -0,0 +1,52 @@
...
+exports[`ABitMoreComplexComponent snapshot test 1`] = `
+<app-a-bit-more-complex
+  anotherVar="false"
+  someVar={[Function Boolean]}
+>
+  
+      <div
+    class="kek"
+  >
+
```
sharikovvladislav: I don't like this formatting. Do you mind adding some format tool like prettier or stuff?

example/src/app/a-bit-more-complex-component/__snapshots__/a-bit-more-complex.component.spec.ts.snap
```diff
@@ -0,0 +1,52 @@
...
+      class="inner-class"
+    >
+      
+          
+      
+      
+          
+      <div
+      >
+        another case
```
sharikovvladislav:
```
      <div
      >
```
is not good. 

Must be:

```
<div>
```

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+describe('ABitMoreComplexComponent', () => {
+  let component: ABitMoreComplexComponent;
+  let fixture: ComponentFixture<ABitMoreComplexComponent>;
+
+  beforeEach(
+    async(() => {
+      TestBed.configureTestingModule({
+        declarations: [ABitMoreComplexComponent, ChildComponent],
+      })
+        .overrideComponent(ChildComponent, {set: {template: 'stubbedBody'}})
```
sharikovvladislav: By the way. Do you mind adding this approach of overriding components to the docs? This is about shallow rendering.

example/src/app/a-bit-more-complex-component/__snapshots__/a-bit-more-complex.component.spec.ts.snap
AhnpGit: Hi, I found what jest use to format their snapshot. It is called pretty-format. Perhaps you can take a look https://www.npmjs.com/package/pretty-format

setupJest.js
```diff
@@ -8,7 +8,34 @@ require('zone.js/dist/sync-test');
 require('zone.js/dist/async-test');
 require('zone.js/dist/fake-async-test');
 require('jest-zone-patch');
+
 const AngularSnapshotSerializer = require('./AngularSnapshotSerializer');
+
+const MySerializer = {
```
AhnpGit: Can you rename this serializer to a more meaningful name ? I think the name "MySerializer" doesn't say what it does to improve the current AngularSnapshotSerializer

sharikovvladislav: Of course, I will. 

This PR is just to have some discussion about how this should be done. @thymikee does not have time for this probably, so this PR is just waiting :(.

example/src/app/a-bit-more-complex-component/__snapshots__/a-bit-more-complex.component.spec.ts.snap
barretodavid: They seem to use [pretty-format](https://github.com/facebook/jest/tree/master/packages/pretty-format#usage-with-plugins) with a custom React plugin

AhnpGit: Hi @sharikovvladislav , @barretodavid,
There is a solution for this. Can you guys please check #86 discussion ?

sharikovvladislav:
@thymikee any thoughts?

thymikee:
I'd rather like this serializer to work on actual values instead of regexing things around. Is there a way you could add this functionality to the current one?

sharikovvladislav:
What do you mean by "actual values"? Do you want me to modify existing serializer? 

Now it is separated. Your serializer doing one thing. Another serializer extends yours by doing something extra. Do you think this approach is wrong?

thymikee:
I think we should strive to have one that does the job well. By actual values I mean testing for objects with certain values as keys (e.g. `_ngcontent`). Also, you're running AngularSnapshotSerializer twice. It should suffice to do it just once.

sharikovvladislav:
okay, as you wish! 

what about general idea? what do you think about it?

iMarv:
If you create a dedicated describe-block for your snapshot tests, you can run `enableProdMode()` before building your test module. this will remove any debugging-directives (like `ng-reflect-*`) and probably result in the snapshot you are seeking

jfcere:
Is this PR going to be merged some day?

I don't really know how and why but after updating to Angular 6 I found that a lot of `_nghost` and `_ngcontent` appeared in the snapshots ... I guess this PR would resolve the issue.

thymikee:
Would love to in some form, but now it looks like the feedback wasn't addressed. Anybody willing to pick this up?

sharikovvladislav:
Hey!

Well, question is still opened.

> what about general idea? what do you think about it? do you think it is good to have this feature in your tool?

I am not sure you need this @thymikee , thats why I don't work on this PR.

thymikee:
Not me, but the project would benefit from stripping some Angular internals.

sharikovvladislav:
So do you mind merging upgraded version of this PR?

This is only one suggestion? https://github.com/thymikee/jest-preset-angular/pull/97#issuecomment-370268820 Anything else? I can fix this.

thymikee:
My concern was https://github.com/thymikee/jest-preset-angular/pull/97#issuecomment-370268820 and actually invoking `AngularSnapshotSerializer` twice

isidro1983:
Hi! I'm following this thread since, it is getting too much hard for me to find a way to get rid of the `ng-reflect-` prefix in my template. I'm currently working on app route testing and need to do some directive filtering. The approach of adding `enableProdMode()` in a separate `describe` block (as @iMarv says [here](https://github.com/thymikee/jest-preset-angular/pull/97#issuecomment-389463330)) makes `DebugElement` no to be available, hence I lose lot of features in order to make my tests work. Does it happen the same to you @iMarv ? 
I've debugged official Angular examples for Route Testing and templates are compiled OK without `ng-reflect-` prefix. (Tests are developed with Karma and Jasmine). I don't see any call to `enableProdMode()` anywhere before running tests.
I tried to be as brief as possible. Sorry if I've made wrong assumptions. I'd be glad to give you more details in case you require them.
Thanks!

iMarv:
@isidro1983 well the ng-reflect stuff should be part of "devMode", so enabling prod mode removes anything related to it.

We are sticking to the fully fledged snapshots now, as they can provide relevant information about what is given to inputs etc

listepo-alterpost:
Any news?

sharikovvladislav:
I will try to implement it this weekend.

пт, 19 апр. 2019 г. в 19:14, listepo-alterpost <notifications@github.com>:

example/src/app/simple/__snapshots__/simple.component.spec.ts.snap
```diff
@@ -4,7 +4,6 @@ exports[`SimpleComponent snapshots 1`] = `
 <app-simple>
   <p>
      simple works!
-
```
sharikovvladislav: Is it okay that I removed that empty lanes from existing snapshots? it is because of added filter function in serializer.

sharikovvladislav:
> Any news?

Done.

sharikovvladislav:
In general: 
- I just added few lines of code into existing angular snapshot serializer.
- I also added a bit complex component example.

@thymikee 
Notice, that this is huge BC and you should bump major version to safely publish this code feature.

Probably, you want to add this behind some configuration flag? What do you think?

ahnpnl:
Angular v8 is almost there, I guess we can merge and hold for release once v8 releases.

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('snapshot test', () => {
+    expect(fixture).toMatchSnapshot();
+  });
+});
```
wtho: The component, test file and the test suite should not just be named "a bit more complex component", we have also other tests than the serializers, so it should definitely explain, that it is a serializer-test.

You can name it something like "complex templating component" or name what is complex about it.

AngularSnapshotSerializer.js
```diff
@@ -29,6 +29,17 @@ const print = (val, print, indent, opts, colors) => {
     .map(node => Array.from(node.renderElement.childNodes).map(print).join(''))
     .join(opts.edgeSpacing);
 
+
+  const nodesWithoutDebugInformation = nodes
+    .split('\n')
```
wtho: What if `opts.edgeSpacing` (see line 30) is not set to `\n`? I think this might not work anymore.

Better make your transformations before the html gets joined.

sharikovvladislav:
@wtho  very good points. And actually my code does not handle such cases it all. Such text will be removed.

> Angular also adds attributes to inner non-component html elments like through directives and animations. I think these are not handled by this serializer, but I might be wrong.

Can you give some examples? Especialy animations.

TODO:
- do not remove _ngcontent/_nghost/ng-reflect text if it is passed as real HTML, but was not generated by Angular and add units for such cases
- we need some configuration flag about new behavior of serializer
- check and handle `class="ng-tns-c1-1` cases

AngularSnapshotSerializer.js
sharikovvladislav: Probably, I can just split and join lines by `opts.edgeSpacing`? What do you think?

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts
sharikovvladislav: Actually, this code was wrote very long time ago, but still... It was "copy-pasted" from other specs in example project.

`example/src/app/calc/calc.component.spec.ts`:

```
  it('should snap', () => {
    expect(fixture).toMatchSnapshot();
  });
```

`example/src/app/simple/simple.component.spec.ts`:
```
  it('snapshots', () => {
    expect(fixture).toMatchSnapshot();
  })
```

And stuff...

I understand your point. Is it critical? If yes, other code have to be refactored too.

sharikovvladislav:
> Another thing that does not seem to be discussed here is class-names in components. They get indexed and each time the test runs in a different order, the indeces are different, e. g. class="ng-tns-c1-1".

Can you give some example of such behavior?

I tried to get this. Here is the example: [click me](https://stackblitz.com/edit/angular-test-class-selectors). I can not see any indexes in class list.

wtho:
I build a small example with a directive and an animation, but luckily both do not add any attributes to the html, they both just add more classes.

https://stackblitz.com/edit/angular-hqvsrr?file=src%2Fapp%2Fapp.component.html

```html
<p _ngcontent-xqo-c12 class="ng-tns-c12-0 test" appsomedirective>
  directive
</p>
<p _ngcontent-xqo-c12 class="ng-tns-c12-0 ng-trigger ng-trigger-triggerName" style>
  animation
</p>
<p _ngcontent-xqo-c12>
  text
</p>
```

Still all children get the same `_ngcontent`-attribute, which should also be handled in some way.

----

The current serializer does not render non-angular component html, it lets `pretty-format` (included in jest) do the job. I think a good solution in the long run would be to use `pretty-format`, but tell it to handle specific attributes different.
Sadly there is no API in `pretty-format` to do that, but we can rewrite the existing `pretty-format`-plugin [`DOMElement.js`](https://github.com/facebook/jest/blob/master/packages/pretty-format/src/plugins/DOMElement.ts) and call `pretty-format` with the plugin to obtain cleaned up html.

This is just a suggestion and open for discussion.

AngularSnapshotSerializer.js
wtho: In line 30 the join is applied. If you use the nodes without joining them there you do not have to split them.

Your join afterwards should use `opts.edgeSpacing` then.

sharikovvladislav:
> the elements inside this module will get the indexed classes. See my example.

So do you suggest to remove such technical (for animations, like `ng-tns-c6-0`) classes in `jest-preset-angular` serializers? What should I do in this PR?

> The current serializer does not render non-angular component html

Sorry, I didn't get. How this paragraph relates to my PR? What exactly you want me to do in this PR?

> we need some configuration flag about new behavior of serializer

Also, lets discuss this option. This have to be done in this PR. We don't want breaking changes after implementing this feature, do we? So this feature has to be hidden behind some configuration flag. By default, this feature will be disabled. If people like this feature, default value of that configuration may be changed in some major version in the future.
Am I right? So I will add some config flag with `false` default value, so this feature would not be applied to existing snapshots in existing projects, which are using `jest-preset-angular`. Also, docs should be updated.

wtho:
> So do you suggest to remove such technical (for animations, like ng-tns-c6-0) classes in jest-preset-angular serializers?

I think we do not have to handle it in this PR directly, but keep it in mind in case some time later it will pop up. Most tests in jest can be run with NOOPAnimations, and I guess NOOP will not add these classes, so it is not so relevant for this PR now.

> Sorry, I didn't get. How this paragraph relates to my PR?

So if you look at the current serializer (or how serializers in jest work), they match elements in the test function and handle these elements if they decide they are responsible for them.
To have a generic way of handling default elements, jest also has built-in serializers, e. g. for json, js and html. To encapsulate this formatting functionality, jest itself provides a package called `pretty-format` (not `prettier` or `prettier-format`). I am not sure how the option `prettierpath-string` plays a role, but probably it uses similar options as the famous formatter `prettier` and can apply them as well.

In our tests, Angular components are formatted by our serializer, but the content in the template is again passed to jest. The inner HTML elements will fallback to use the `pretty-format`, as they do not match the `test`-fn in the serializer. If inner elements are Angular components as well, they will again be matched and serialized by the serializer you write.

The problem that arises here is that the serializer is not in charge of elements like `<p>` or `<div>` and cannot skip the rendering `ng-content` attributes on them. My suggestion was we would supply a serializer for these basic html elements as well, which means more work though.

> we need some configuration flag about new behavior of serializer

I think we just create a new serializer for the start. Which serializer should be used can be easily configured in the jest config. So if anyone is interested in trying out this new serializer, they can simply plug it in instead of the other one. Whenever we and the users want to switch to the new one permanently, we can just switch them, but still provide the old one as an option.
I could find no runtime-argument-based configuration in jest serializers to change the serializing behavior, maybe @thymikee can confirm this.

wtho:
Ok so I might not have explained well enough, why I am talking about `pretty-format`.

The current approach in your implementation is letting the AngularSerializer and `pretty-format` do the serialization of the components to html and then modify it using regexes. It just feels cleaner to already intervene at the serialization, to avoid any serialization of the unwanted attributes than handling the string-based representation afterwards.
And this would mean we have to change/replace the serialization, that `pretty-format` does.

sharikovvladislav:
> The problem that arises here is that the serializer is not in charge of elements like <p> or <div> and cannot skip the rendering ng-content attributes on them. My suggestion was we would supply a serializer for these basic html elements as well, which means more work though.

Do you mean [this](https://stackblitz.com/edit/angular-1vsw64) situation? Yeah, I understood.

According to all of this information it is really better not to change AngularSnapshotSerializer, but implement another serializer, which will trim specific attributes (like nghost, ngcontent, ngreflectif and stuff). Looks like it would be better solution. Also, all logic about trimming this stuff would be extracted to separate serializer. 

> I think we just create a new serializer for the start. Which serializer should be used can be easily configured in the jest config.

Actually, looks like we almost got final solution for the problem. If this work (removing debug attributes) would perform new serializer, config flag is not needed. It would be handled by specifying serializers array.

So, the plan is: create new serializer, which will do the work and add it *before* AngularSnapshotSerializer. Test function will get only nodes which has needed attributes. Print function will remove that attributes. Am I right?

sharikovvladislav:
> So, the plan is

@wtho @ahnpnl so is it ok ? Or do you have something to add?

ahnpnl:
I like that idea 👍

wtho:
Yeah!

listepo-alterpost:
@sharikovvladislav can we remove and `ng-version`?

sharikovvladislav:
> hm

How do you create component fixture so ng-version comes to your snapshot?

I thought it comes somewhere on app root element... Do you test app root component?

Technically, yes, I think we can remove anything.

sharikovvladislav:
@wtho @ahnpnl do you mind using `pretty-format` plugin in `jest-preset-angular`?

Suggestion:
1. add `pretty-format` package
2. import it in my new serializer
3. implement test & serialize functions.

Test will be test from prettyFormat.plugins.DOMElement + check if node contain angular specific attribute and serialize will do this work:
1. remove angular specific attributes from node
2. call serialize function from DOMElement plugin

Here is sourcecode of DOMElement plugin: https://github.com/facebook/jest/blob/master/packages/pretty-format/src/plugins/DOMElement.ts

Reason: do not implement something which is already implemented in internal jest plugins.

Profit: 
- less code in serializer
- no influence on node_modules dir size since pretty-format is already dependency of current packages list

Thoughts? @wtho @ahnpnl @thymikee

wtho:
I think this is great.

As you said Jest already includes `pretty-format` anyway.

Some thoughts about the attribute removal: I am not sure though about it. In case the component nodes are used in other `expect`s, we should not modify them. On the other hand, working on a copy might slow down the tests.

But managing a copy of `DOMElement.ts` is also a tedious long-term task, that we would not want to commit to in this preset.

Best would be to have some hooks in `DOMElement.ts` or to be able to pass a shallow copy of each DOM node to the `DOMElement.ts`, but I have not looked too deep into this.

sharikovvladislav:
@wtho Ok, I pushed something.

I just want you to check general idea.

What is wrong now:
1. new serializer is enabled by default
2. no `pretty-format` as dependency of `jest-preset-angular`
2. no docs

sharikovvladislav:
OMG.... CI is so fast.... Thats really perfect. Cool.

Looks like, my serializer added some empty lanes... didn't it?

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts
wtho: I see. Still I think "SimpleComponent" alone is clear what to expect in the component. "ABitMoreComplexComponent" will always only stand in comparison to "SimpleComponent", which is not as good.

I would recommend e. g. "MediumComplexComponent"

AngularSpecificAttributesRemover.js
```diff
@@ -0,0 +1,46 @@
...
+  isElement(val) && jestDOMElementSerializer.test(val) && serializeTestFn(val);
+
+module.exports = {
+  serialize: serialize,
+  test: test
+};
+
+function serializeTestFn(val) {
+  return Object.keys(val.attributes)
+    .map(key => val.attributes[key].name)
```
wtho: You can direclty use `Object.values(...).map(value => value.name)` here, as you do not really care for the key.

AngularSpecificAttributesRemover.js
```diff
@@ -0,0 +1,46 @@
...
+  Object.keys(nodeCopy.attributes)
+    .map(key => nodeCopy.attributes[key].name)
+    .filter(attributeName => attributeName.includes('ng-reflect'))
+    .forEach(attributeNameToRemove => nodeCopy.attributes.removeNamedItem(attributeNameToRemove));
+
+  return jestDOMElementSerializer.serialize(nodeCopy, ...rest);
+};
+
+const test = val =>
+  isElement(val) && jestDOMElementSerializer.test(val) && serializeTestFn(val);
```
wtho: Is `isElement` really required or should the test function from `DOMElement` not be enough already?

example/src/app/a-bit-more-complex-component/child.component.less
```diff
@@ -0,0 +1,5 @@
+.kek { color: red }
+
+:host {
+  color: blue;
+}
```
wtho: Is this file really required? Let's keep the test as simple as possible. Plus I think we do not have a preprocessor configured.

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.less
```diff
@@ -0,0 +1,5 @@
+.kek { color: red }
+
+:host span {
+  color: red;
+}
```
wtho: Is this file really required? Let's keep the test as simple as possible. Plus I think we do not have a preprocessor configured.

example/src/app/ng-reflect-as-text/ng-reflect-as-text.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'app-ng-reflect-as-text',
+  templateUrl: './ng-reflect-as-text.component.html',
+  styleUrls: ['./ng-reflect-as-text.component.css']
```
wtho: If the `styleUrls` file is empty, we can just skip this part and remove the file

sharikovvladislav:
> @sharikovvladislav can we remove and `ng-version`?

Can you create simple example with `ng-version` and publish it somewhere? I can not reproduce it in `example` app in that project. Very strange...

@wtho any thoughts why I don't get `nghost`/`ngcontent`/`ng-version` attributes?

example/src/app/ng-reflect-as-text/ng-reflect-as-text.component.ts
sharikovvladislav: Ok. I also removed empty css files from other components.

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.less
sharikovvladislav: I removed it.

example/src/app/a-bit-more-complex-component/child.component.less
sharikovvladislav: I removed it. It was try to get nghost/ngcontent attributes in snapshots.

AngularSpecificAttributesRemover.js
sharikovvladislav: Done

AngularSpecificAttributesRemover.js
sharikovvladislav: Hm... I think I got some problem yesterday without it.... but now everything works okay... Looks like it is not needed. I removed it.

sharikovvladislav:
Actually... we could implement something like this...

1. New serializer with basic removes like ng-reflect/nghost/ngcontent
2. Ability to provide custom regexps to remove custom stuff from snapshots (they will be user defined) and if anybody wants to remove something special he can write regexp for it and provide it as jest-preset config.

@wtho what do you think? Maybe in future PRs...

AngularSpecificAttributesRemover.js
sharikovvladislav: It was for case when val.attributes is undefined.

I added check for it and now everything is okay.

sharikovvladislav:
TODO:
- [ ] fix docs
- [ ] find case about nghost/ngcontent/ngversion attributes and remove that attributes (now only ng-reflect remover is implemented)

sharikovvladislav:
Looks like I have problem, folks @wtho @ahnpnl 

On your CI (there) I have this:
image

And locally I have this: 
image
]

Locally I have this environment:
image

listepo-alterpost:
@sharikovvladislav sure, https://github.com/listepo-alterpost/jestTestApp/blob/master/libs/mylib/src/lib/test/__snapshots__/test.component.spec.ts.snap#L6

sharikovvladislav:
> @sharikovvladislav sure, https://github.com/listepo-alterpost/jestTestApp/blob/master/libs/mylib/src/lib/test/__snapshots__/test.component.spec.ts.snap#L6

What you want to achieve by using this construction?

```
expect(fixture.nativeElement).toMatchSnapshot();
```

Why don't you using

```
expect(fixture).toMatchSnapshot();
```

Just wondering.

listepo-alterpost:
@sharikovvladislav my bad, thanks

wtho:
> Looks like I have problem
I ran the tests locally and they also pass.
It seems like the ci uses jest-cache. We should definitely avoid that on ci machines.

To disable the cache add `--no-cache` in the `test:ci` commands in both `package.json`.

Rerunning tests should not have the newline problem anymore.

AngularSpecificAttributesRemover.js
```diff
@@ -0,0 +1,27 @@
...
+
+module.exports = {
+  serialize: serialize,
+  test: test
+};
+
+function serializeTestFn(val) {
+  return val.attributes !== undefined && Object.values(val.attributes)
+    .map(attribute => attribute.name)
+    .some(attributeName => attributeName.includes('ng-reflect'))
```
wtho: I think `attributeName.startsWith` is more appropriate here, Angular's naming convention is that Angular specific attributes start with `ng`. Imagine someone having an attribute called `some-wrong-reflect-attribute`.

example/src/app/ng-reflect-as-text/ng-reflect-as-text.component.ts
```diff
@@ -0,0 +1,14 @@
...
+@Component({
+  selector: 'app-ng-reflect-as-text',
+  templateUrl: './ng-reflect-as-text.component.html'
+})
+export class NgReflectAsTextComponent implements OnInit {
+
+  constructor() { }
+
+  ngOnInit() {
+  }
```
wtho: I know, it comes preconfigured like this, but I like it minimal. Let's strip off the constructor and remove `OnInit` here, so later the code becomes more clear what it is about.

jest-preset.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
   },
   transformIgnorePatterns: ['node_modules/(?!@ngrx)'],
   snapshotSerializers: [
+    'jest-preset-angular/AngularSpecificAttributesRemover.js',
```
wtho: Do you mind adding the suffix `Serializer` here?

Maybe let's also think about the naming of the serializer. It serializes Angular components as HTML, skipping `ng-*`-attributes. My suggestion therefore would not be to focus on the removal (which is the process the serializer does), but on the result, e. g. something like `AngularNoNgAttributesSnapshotSerializer.js

example/src/app/ng-reflect-as-text/ng-reflect-as-text.component.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(NgReflectAsTextComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
wtho: I think this test is obsolete. We have it in many other tests already. This will save us some time.

example/src/app/ng-reflect-as-text/ng-reflect-as-text.component.html
```diff
@@ -0,0 +1,3 @@
+<p>
+  ng-reflect-as-text works!
```
wtho: Do you mind adding `ng-reflect` in an attribute value here? Like `<p some-attr="ng-reflect-as-attr-arg works">`

AngularSpecificAttributesRemover.js
sharikovvladislav: Done

example/src/app/ng-reflect-as-text/ng-reflect-as-text.component.ts
sharikovvladislav: Done

jest-preset.js
sharikovvladislav: Done

example/src/app/ng-reflect-as-text/ng-reflect-as-text.component.spec.ts
sharikovvladislav: Done

example/src/app/ng-reflect-as-text/ng-reflect-as-text.component.html
sharikovvladislav: Done

sharikovvladislav:
> TODO:
> 
> * [ ]  fix docs
> * [ ]  find case about nghost/ngcontent/ngversion attributes and remove that attributes (now only ng-reflect remover is implemented)


I fixed all points. Note that this is till not finished.

And yes, we still have problem on CI build.

wtho:
Ok we have a problem on the CI side.

I SSHed into one of the builds and realized yarn installs an old version of `jest-preset-angular` in the example app `node_modules`. It cached the older build and reuses it, although changes have happened in the preset.

I hoped yarn would take a hash of the project to ensure the preset being up-to-date, but this is not the case.

Let's remove the preset manually before installing as a workaround:

`.circleci/config.yaml`
```yaml
  [...]
  install-test-example:
    executor: node-10
    steps:
      - attach_workspace:
          at: .
      - restore_cache:
          keys:
            - yarn-cache-example-{{ .Branch }}-{{ checksum "yarn.lock" }}
            - yarn-cache-example-{{ .Branch }}-
      - run:
          name: Install Example Dependencies
          command: |
            cd example
+           rm -rf node_modules/jest-preset-angular
            yarn install --frozen-lockfile
      - save_cache:
  [...]
```

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.ts
```diff
@@ -0,0 +1,28 @@
...
+        <div>another case</div>
+        <span>some html</span>
+      </div>
+      <div>one more case case</div>
+      <div><app-child-component kek="someVar"></app-child-component></div>
+    </div>
+  `,
+  styles: [`
+    .kek { color: red }
+  `]
```
wtho: I still think we do not need the styles at all. No one ever looks at the components and they just add noise.

The browser/jest does not complain if there is not an actual class with the name `kek` loaded.

sharikovvladislav: Actually.... We need styles or style urls in case of that fact. If styles are provided to the component they must be encapsulated (by default). I don't know why, but even that fact that I specify style (inline in component selector) or styleUrls - these attributes nghost/ngcontent are not created for some reasons. Looks like there is come code that removes them....

My initial code with style was about that. I don't understand why nghost/ngcontent attributes are not in snapshots now.

sharikovvladislav:
> Let's remove the preset manually before installing as a workaround:

Do you want something from me or will you take this problem?

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.ts
sharikovvladislav: Will we handle nghost/ngcontent in this PR?

Or do you understand why this attributes are not in result snapshot?

wtho:
I pushed the workaround

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.ts
wtho: Good point.
I have no clue why they disappeared. Let's keep investigating a bit more.

wtho: I know why. We strip the styles in the preset, as we do not care about them in jest tests.

But there are several ways to avoid the stripping. This one should work (though I haven't tested it):
```ts
 @Component({
  selector: 'app-medium-complex',
  template: `...`,
  ['styles']: [`
    .kek { color: red }
  `]
})
```

sharikovvladislav: Ok. In that case... should I handle nghost/ngcontent if you trim styles anyway?

By the way.... Where is that code what removes styles from the components?

wtho: Yeah, let's handle them just in case.

sharikovvladislav: Also, what about ng-version? I don't have code for trimming version at the moment. But still somebody got ng-version (this parameter is dynamic and snapshots mustn't relay on such parameter without doubts). @listepo-alterpost had that problem for example.

wtho: I think `ng-version` is only added to the root app, I *guess* it is only rendered when Angular is bootstrapped with a root module/component, not using `TestBed` (might be wrong about it).
I do not see why someone would serialize a whole app using jest, so I do not think this is required, but go ahead if you like to add it.

sharikovvladislav: > 

Not really. Change snapshot test in simple component to `expect(fixture.nativeElement).toMatchSnapshot();`. You will get this stuff:

image

SimpleComponent is not root component, but still it gets ng-version attribute. I think that in in test environment.

Also, here is full dom snapshot starting from body tag:

```
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`SimpleComponent snapshots 1`] = `
<body>
  <div
    id="root1"
    ng-version="7.2.8"
  >
    <p>
       simple works!

    </p>
  </div>
</body>
`;
```

I tried `expect(document.querySelector('body')).toMatchSnapshot();` this expect.

So, if anybody will snapshot full body (which is strange) or anybody fill snapshot not fixture, but fixture.nativeElement (more real, as I think) he will get this problem.

> Edit: we can still add it later if it gets requested.

It might be breaking change after we initially merge this to the project. So better to implement this now, if it is possible.

So now: 

- I add example with styles and `ngcontent`/`nghost` properties and I will add code to new serializer to remove them (despite of `jest-preset-angular` removes styles anyway)
- I remove `ng-version` attributes too
- I need fix docs (or don't?). We need some discussion I think. If I understand correcly, you don't want this serializer by default. So I have to add section how to remove angular specific attributes by adding new serializer. Can you clarify how this will work in general, so I can describe it in docs?

wtho: Ok cool! I was not aware how it can be reproduced.

> I add example with styles and ngcontent/nghost properties
> I remove ng-version attributes too

Yeah, we need tests for this. But maybe we can reuse one of the components already created.

> I need fix docs

The serializer is currently configured in 
https://github.com/thymikee/jest-preset-angular/tree/master/jest-preset.js. We can add a commented out line `// 'jest-preset-angular/AngularSpecificAttributesRemover.js',`, but leave it in there.
We should also put it in the exposed configuration in `README.md` and add some lines about the new serializer being available.
As soon as `jest-preset-angular` will release v8, I would like to propose to use this serializer by default, then we can document it as breaking change as well.
This way we can already publish it and users can already try it out.
The old serializer will still be available for users that prefer the old serialization. 

So these steps are required for the docs:
* Add short section in `README.md`, a bullet point in https://github.com/thymikee/jest-preset-angular#brief-explanation-of-config and the addition of the commented out serializer is sufficient
* A short description in `CHANGELOG.md` including a quick migration guide

sharikovvladislav: KK, I need some time. Will be done todays evening.

sharikovvladislav: Hm, what section in readme describes how to rewrite default `jest-preset.js` configuration?

I do not understand how end-user can rewrite `snapshotSerializer` settings. Could you explain this please?

We need to decide what we want in example app?
- snapshots with angular-specific attributes (since new serializer is commented by default)
- clearer snapshots without such attributes

How to rewritet this setting ? o_O

sharikovvladislav:
I added some docs & changelog. What do you think? My english is not perfect, so feel free to fix it if you think it is neccessary. 

One point which I don't understand... What section in readme describes how to rewrite default `jest-preset.js` configuration?
I do not understand how end-user can rewrite `snapshotSerializer` settings. Could you explain this please?

We need to decide what we want in example app?
- snapshots with angular-specific attributes (since new serializer is commented by default)
- clearer snapshots without such attributes (so I have to rewrite default setting in example app)

How to rewrite this setting ? o_O

ahnpnl:
I think there is none, but you can add a new section right below Expose Configuration 😀

sharikovvladislav:
Mhm... Ye, I understand this... it is not the problem.

How to redefine `jest-preset.js` preset in your project? This is my question. I can not find anything about it in `jest-preset-angular` docs.

I found this way:

1. Create any js file somewhere in your project, for example `my-preset`.
2. Require original preset in it and now you can redefine its options.
3. Specify this file in package.json in the `jest.preset` property. Default is `jest-preset-angular`.

`my-preset` contents should be this for example:

```
const originalPreset = require('jest-preset-angular/jest-preset');
const redefinedPreset = { ...originalPreset };

redefinedPreset.snapshotSerializers = [
  'jest-preset-angular/AngularSpecificAttributesRemover',
  ...originalPreset.snapshotSerializers
];

module.exports = redefinedPreset;
```

Now jest will use `my-preset` when runs.

I found this way. Is it obvious? o_O. If yes, okay, I will just add some information like "you should redefine". But I think it is better to specify how to redefine original `jest-preset-angular` props.

What do you think?

- should I just add section about "you need redefine snapshotSerializers option"
- should I add point above + add example how you can redefine original configuration (I can't find anything on how to redefine original pacakge config).

?

wtho:
Almost, only that they do not have to define a preset. They can override the snapshotSerializers directly in their `jest.config.js`.

sharikovvladislav:
Ehm.. sorry... Is it obvious? I am not pro guy in jest.... I don't understand.

This is what I get...

Case: `jest.config.js` has such contents:
```
module.exports = {
  snapshotSerializers: [
    'jest-preset-angular/AngularSpecificAttributesRemover.js',
    'jest-preset-angular/AngularSnapshotSerializer.js',
    'jest-preset-angular/HTMLCommentSerializer.js',
  ],
};
```

I get something like this:
image

Looks like `ts-jest` is not applied.

If I put whole jest-preset.js and rewrite snapshotSerializer. So this is how `jest.config.js` looks like: 

```
module.exports = {
  globals: {
    'ts-jest': {
      tsConfig: '<rootDir>/src/tsconfig.spec.json',
      stringifyContentPathRegex: '\\.html$',
      astTransformers: [require.resolve('jest-preset-angular/InlineHtmlStripStylesTransformer')],
    },
  },
  transform: {
    '^.+\\.(ts|js|html)$': 'ts-jest',
  },
  testEnvironment: 'jest-environment-jsdom-thirteen',
  moduleFileExtensions: ['ts', 'html', 'js', 'json'],
  moduleNameMapper: {
    '^src/(.*)$': '<rootDir>/src/$1',
    '^app/(.*)$': '<rootDir>/src/app/$1',
    '^assets/(.*)$': '<rootDir>/src/assets/$1',
    '^environments/(.*)$': '<rootDir>/src/environments/$1',
  },
  transformIgnorePatterns: ['node_modules/(?!@ngrx)'],
  snapshotSerializers: [
    'jest-preset-angular/AngularSpecificAttributesRemover.js',
    'jest-preset-angular/AngularSnapshotSerializer.js',
    'jest-preset-angular/HTMLCommentSerializer.js',
  ],
};
```

I get undefined in each snapshot, like this:
image

Can you give some example of how to redefine ONLY snapshotSerializers? I don't even understand how to google it. Sry:D

Ehm is it really obvious and it is only mine problem ? :D

wtho:
No problem! Always ask if someone is unclear.

As documented on [jestjs.io](https://jestjs.io/docs/en/configuration#preset-string) or done in the [example app](https://github.com/thymikee/jest-preset-angular/blob/master/example/package.json#L46), you can reference a preset in your configuration. Jest will then take all these options, except of the ones you override directly in the config.

`jest.config.js`
```js
module.exports = {
  preset: 'jest-preset-angular',
  snapshotSerializers: [
    // ...
  ]
}
```

sharikovvladislav:
Hm.

If I add `jest.config.js` with that contents: 

```
module.exports = {
  preset: 'jest-preset-angular'
};
```

I get this output of `jest` command:

```
~/git/jest-preset-angular/example   master ●✚  jest
 FAIL  src/app/simple/simple.component.spec.ts
  ● Test suite failed to run

    Cannot find module '@lib/testing' from 'simple.component.spec.ts'

      1 | import { async, ComponentFixture } from '@angular/core/testing';
      2 | 
    > 3 | import {ConfigureFn, configureTests} from '@lib/testing';
        | ^
      4 | 
      5 | import { SimpleComponent } from './simple.component';
      6 | 

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:203:17)
      at Object.<anonymous> (src/app/simple/simple.component.spec.ts:3:1)

 FAIL  src/app/on-push/on-push.component.spec.ts
  ● Test suite failed to run

    Cannot find module '@lib/testing' from 'on-push.component.spec.ts'

      2 | import { ChangeDetectionStrategy } from '@angular/core';
      3 | 
    > 4 | import { ConfigureFn, configureTests } from '@lib/testing';
        | ^
      5 | 
      6 | import { OnPushComponent } from './on-push.component';
      7 | 

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:203:17)
      at Object.<anonymous> (src/app/on-push/on-push.component.spec.ts:4:1)

 FAIL  src/app/heroes/heroes.component.spec.ts
  ● Test suite failed to run

    Cannot find module '@lib/testing' from 'heroes.component.spec.ts'

      1 | import {async, ComponentFixture} from '@angular/core/testing';
      2 | import {HeroesComponent} from './heroes.component';
    > 3 | import {ConfigureFn, configureTests} from '@lib/testing';
        | ^
      4 | import {Hero} from '../service/hero';
      5 | import {HeroService} from '../service/hero.service';
      6 | import {Subject} from 'rxjs';

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:203:17)
      at Object.<anonymous> (src/app/heroes/heroes.component.spec.ts:3:1)

 FAIL  src/app/app.component.spec.ts
  ● Test suite failed to run

    Cannot find module '@lib/testing' from 'app.component.spec.ts'

       8 | import { NO_ERRORS_SCHEMA } from '@angular/core';
       9 | 
    > 10 | import { ConfigureFn, configureTests } from '@lib/testing';
         | ^
      11 | 
      12 | import { AppComponent } from './app.component';
      13 | 

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:203:17)
      at Object.<anonymous> (src/app/app.component.spec.ts:10:1)

 FAIL  src/app/calc/calc.component.spec.ts
  ● Test suite failed to run

    Cannot find module '@lib/testing' from 'calc.component.spec.ts'

      1 | import { async, ComponentFixture, TestBed } from '@angular/core/testing';
      2 | 
    > 3 | import { configureTests, ConfigureFn } from '@lib/testing';
        | ^
      4 | 
      5 | import { CalcComponent } from 'app/calc/calc.component';
      6 | 

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:203:17)
      at Object.<anonymous> (src/app/calc/calc.component.spec.ts:3:1)

 FAIL  src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts (5.287s)
  ● ABitMoreComplexComponent › should create

    Failed: "Zone is needed for the async() test helper but could not be found. Please make sure that your environment includes zone.js/dist/zone.js"

       9 |   let fixture: ComponentFixture<ABitMoreComplexComponent>;
      10 | 
    > 11 |   beforeEach(
         |   ^
      12 |     async(() => {
      13 |       TestBed.configureTestingModule({
      14 |         declarations: [ABitMoreComplexComponent, ChildComponent],

      at Env.beforeEach (node_modules/jest-jasmine2/build/jasmineAsyncInstall.js:41:24)
      at Suite.<anonymous> (src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts:11:3)
      at Object.<anonymous> (src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts:7:1)

  ● ABitMoreComplexComponent › should create

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.createComponent (../../packages/core/testing/src/test_bed.ts:594:10)
      at Function.TestBedViewEngine.createComponent (../../packages/core/testing/src/test_bed.ts:247:36)
      at Object.<anonymous> (src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts:22:23)

  ● ABitMoreComplexComponent › should create

    expect(received).toBeTruthy()

    Received: undefined

      26 | 
      27 |   it('should create', () => {
    > 28 |     expect(component).toBeTruthy();
         |                       ^
      29 |   });
      30 | 
      31 |   it('snapshot test', () => {

      at Object.<anonymous> (src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts:28:23)

  ● ABitMoreComplexComponent › snapshot test

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot "ABitMoreComplexComponent snapshot test 1".

    - Snapshot
    + Received

    - <app-a-bit-more-complex
    -   anotherVar="false"
    -   someVar={[Function Boolean]}
    - >
    -   <div
    -     class="some-wrapper-class"
    -   >
    -      diary works! 
    -     
    -     <div
    -       class="inner-class"
    -     >
    -       
    -       <div>
    -         another case
    -       </div>
    -       <span>
    -         some html
    -       </span>
    -     </div>
    -     <div>
    -       one more case case
    -     </div>
    -     <div>
    -       <app-child-component
    -         someinput="someVar"
    -       >
    -         stubbedBody
    -       </app-child-component>
    -     </div>
    -   </div>
    - </app-a-bit-more-complex>
    + undefined

      30 | 
      31 |   it('snapshot test', () => {
    > 32 |     expect(fixture).toMatchSnapshot();
         |                     ^
      33 |   });
      34 | });
      35 | 

      at Object.<anonymous> (src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts:32:21)

 › 1 snapshot failed.
 FAIL  src/app/simple-with-styles/simple-with-styles.component.spec.ts (5.307s)
  ● SimpleWithStylesComponent › should generate snapshot without nghost/ngcontent

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot "SimpleWithStylesComponent should generate snapshot without nghost/ngcontent 1".

    - Snapshot
    + Received

    - <app-simple-with-styles>
    -   <p>
    -      simple-with-styles works!
    -   
    -   </p><div
    -     class="some-class"
    -   >
    -     <app-child-component>
    -       <div
    -         class="child"
    -       >
    -           
    -         <div>
    -            rly 
    -           <p>
    -             complex
    -           </p>
    -           <div>
    -             component 
    -             <div>
    -               oh my god!
    -             </div>
    -           </div>
    -         </div>
    -       </div>
    -     </app-child-component>
    -   </div>
    - </app-simple-with-styles>
    + undefined

      22 | 
      23 |   it('should generate snapshot without nghost/ngcontent', () => {
    > 24 |     expect(fixture).toMatchSnapshot();
         |                     ^
      25 |   });
      26 | });
      27 | 

      at Object.<anonymous> (src/app/simple-with-styles/simple-with-styles.component.spec.ts:24:21)

 › 1 snapshot failed.
 FAIL  src/app/ng-reflect-as-text/ng-reflect-as-text.component.spec.ts (5.296s)
  ● NgReflectAsTextComponent › snapshots

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot "NgReflectAsTextComponent snapshots 1".

    - Snapshot
    + Received

    - <app-ng-reflect-as-text>
    -   <p
    -     data-some-attr="ng-reflect-as-attr-arg works"
    -   >
    -      ng-reflect-as-text works!
    -   
    -   </p>
    - </app-ng-reflect-as-text>
    + undefined

      21 | 
      22 |   it('snapshots', () => {
    > 23 |     expect(fixture).toMatchSnapshot();
         |                     ^
      24 |   });
      25 | });
      26 | 

      at Object.<anonymous> (src/app/ng-reflect-as-text/ng-reflect-as-text.component.spec.ts:23:21)

 › 1 snapshot failed.
 FAIL  src/app/service/hero.service.spec.ts (5.325s)
  ● Service: HeroService › should create an instance successfully

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:473:10)
      at Function.TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:243:36)
      at Object.<anonymous> (src/app/service/hero.service.spec.ts:38:23)

  ● Service: HeroService › should create an instance successfully

    expect(received).toBeDefined()

    Received: undefined

      49 | 
      50 |   it('should create an instance successfully', () => {
    > 51 |     expect(service).toBeDefined()
         |                     ^
      52 |   })
      53 | 
      54 |   it('should call the GET heroes api and return all results', () => {

      at Object.<anonymous> (src/app/service/hero.service.spec.ts:51:21)

  ● Service: HeroService › should create an instance successfully

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.execute (../../packages/core/testing/src/test_bed.ts:484:10)
      at Object.<anonymous> (../../packages/core/testing/src/test_bed.ts:690:40)

  ● Service: HeroService › should call the GET heroes api and return all results

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:473:10)
      at Function.TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:243:36)
      at Object.<anonymous> (src/app/service/hero.service.spec.ts:38:23)

  ● Service: HeroService › should call the GET heroes api and return all results

    TypeError: Cannot read property 'getHeroes' of undefined

      55 |     let actualDataAll = {}
      56 | 
    > 57 |     service.getHeroes().subscribe(data => actualDataAll = data)
         |             ^
      58 | 
      59 |     backend.expectOne((req: HttpRequest<any>) => {
      60 |       return req.url === `${heroesUrl}`

      at Object.<anonymous> (src/app/service/hero.service.spec.ts:57:13)

  ● Service: HeroService › should call the GET heroes api and return all results

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.execute (../../packages/core/testing/src/test_bed.ts:484:10)
      at Object.<anonymous> (../../packages/core/testing/src/test_bed.ts:690:40)

  ● Service: HeroService › should call the GET hero api and return the result

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:473:10)
      at Function.TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:243:36)
      at Object.<anonymous> (src/app/service/hero.service.spec.ts:38:23)

  ● Service: HeroService › should call the GET hero api and return the result

    TypeError: Cannot read property 'getHero' of undefined

      69 |     let actualData = {}
      70 | 
    > 71 |     service.getHero(1).subscribe(data => actualData = data)
         |             ^
      72 | 
      73 |     backend.expectOne((req: HttpRequest<any>) => {
      74 |       return req.url === `${heroesUrl}`

      at Object.<anonymous> (src/app/service/hero.service.spec.ts:71:13)

  ● Service: HeroService › should call the GET hero api and return the result

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.execute (../../packages/core/testing/src/test_bed.ts:484:10)
      at Object.<anonymous> (../../packages/core/testing/src/test_bed.ts:690:40)

  ● Service: HeroService › should send an expected GET request and throw error to console when an error occurs

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:473:10)
      at Function.TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:243:36)
      at Object.<anonymous> (src/app/service/hero.service.spec.ts:38:23)

  ● Service: HeroService › should send an expected GET request and throw error to console when an error occurs

    TypeError: Cannot read property 'getHero' of undefined

      82 | 
      83 |   it('should send an expected GET request and throw error to console when an error occurs', () => {
    > 84 |     service.getHero(1).subscribe()
         |             ^
      85 | 
      86 |     const getHeroRequest = backend.expectOne((req: HttpRequest<any>) => {
      87 |       return req.url === `${heroesUrl}`

      at Object.<anonymous> (src/app/service/hero.service.spec.ts:84:13)

  ● Service: HeroService › should send an expected GET request and throw error to console when an error occurs

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.execute (../../packages/core/testing/src/test_bed.ts:484:10)
      at Object.<anonymous> (../../packages/core/testing/src/test_bed.ts:690:40)

  ● Service: HeroService › should return an observable of undefined and print error to console

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:473:10)
      at Function.TestBedViewEngine.get (../../packages/core/testing/src/test_bed.ts:243:36)
      at Object.<anonymous> (src/app/service/hero.service.spec.ts:38:23)

  ● Service: HeroService › should return an observable of undefined and print error to console

    TypeError: Cannot read property 'handleError' of undefined

       97 | 
       98 |   it('should return an observable of undefined and print error to console', () => {
    >  99 |     const result = service.handleError(new HttpErrorResponse({ error: 'Error occurs' }), 'test method')
          |                            ^
      100 | 
      101 |     expect(console.error).toHaveBeenCalled()
      102 |     result.subscribe(value => expect(value).toBeUndefined())

      at Object.<anonymous> (src/app/service/hero.service.spec.ts:99:28)

  ● Service: HeroService › should return an observable of undefined and print error to console

    TypeError: Cannot read property 'getComponentFromError' of null

      at TestBedViewEngine._initIfNeeded (../../packages/core/testing/src/test_bed.ts:393:46)
      at TestBedViewEngine.execute (../../packages/core/testing/src/test_bed.ts:484:10)
      at Object.<anonymous> (../../packages/core/testing/src/test_bed.ts:690:40)

Snapshot Summary
 › 3 snapshots failed from 3 test suites. Inspect your code changes or re-run jest with `-u` to update them.

Test Suites: 9 failed, 9 total
Tests:       9 failed, 9 total
Snapshots:   3 failed, 3 total
Time:        11.547s
Ran all test suites.
 ✘  ~/git/jest-preset-angular/example   master ●✚  
```

sharikovvladislav:
Eeeeehm... Ok.

2 ways: 

1. define snapshot serializers array in `package.json`
2. define `jest.config.js` and run NOT `jest` command, but `jest --config jest.config.js` and define full config from current `package.json` file + `snapshotSerializers` option in it. 

Any of that way worked fine.

sharikovvladislav:
Ok. Looks like now everything is finished. 

What do you think? @ahnpnl @wtho?

AngularNoNgAttributesSnapshotSerializer.js
```diff
@@ -0,0 +1,35 @@
...
+
+const test = val =>
+  jestDOMElementSerializer.test(val) && serializeTestFn(val);
+
+module.exports = {
+  serialize: serialize,
+  test: test
+};
+
+function serializeTestFn(val) {
```
ahnpnl: Can this use function expression ?

sharikovvladislav: Done

package.json
```diff
@@ -8,6 +8,7 @@
   "license": "MIT",
   "dependencies": {
     "jest-environment-jsdom-thirteen": "^1.0.0",
+    "pretty-format": "^24.0.0",
```
ahnpnl: I remember when installing `jest`, it will automatically install `pretty-format` too. I'm not sure but would you please confirm it ? If `jest` installs for us, we don't need to add in our `dependencies`

sharikovvladislav: Well, maybe, but... But in some future `pretty-format` could be removed from our dependency (probably `jest`) and this project will get error because there is no `pretty-format` in `node_modules`. 

So this explicit dependency will prevent errors in such case.

Anyway, `node_modules` size is same. Even `yarn.lock` was not changed.

Only one problem: we have to update version of `pretty-format` too.

sharikovvladislav: So I am not sure it is good idea to have implicit dependency. If you force your opinion - sure I will remove it.

sharikovvladislav:
TODO: 
- [ ] check backward compatibility

sharikovvladislav:
Ok, I have project where I started using jest + this tool and I checked new `jest-preset-angular` on this project. Looks like everything works fine.

package.json
wtho: I prefer having it as a dependency. Npm/Yarn are smart enough to install the package only once. Plus we will be notified by greenkeeper when the dependency updates and have more control in case pretty-format changes it API.

wtho:
@sharikovvladislav one more thing, I really would prefer `a-bit-more-complex.component` renamed to `complex.component` or `medium-complex.component`, to not be relative to another component, but have a valid standalone name.

If you do not like this naming change, I am ok with it, but I just did not want to leave this undiscussed.

sharikovvladislav:
> @sharikovvladislav one more thing, I really would prefer `a-bit-more-complex.component` renamed to `complex.component` or `medium-complex.component`, to not be relative to another component, but have a valid standalone name.
> 
> If you do not like this naming change, I am ok with it, but I just did not want to leave this undiscussed.

Done.

ahnpnl:
i think all good to go

wtho:
@thymikee could you please have a look?

By putting it as an optional serializer for now we can immediately publish a new minor version and do not have to hold it back.
Later on release of v8 we can use it as the main serializer, as several people preferred the cleaner html without angular debug properties.
Opting out will still be easy, as it does not touch the old serializer.

---

Also this PR fixes some issue with CI that we did not stumble upon before: we cache `example/node_modules` to speed up CI, but this leads to `jest-preset-angular` in `example/node_modules` to be cached and reused, always the same version whichever is currently in the cache (which is no issue for yarn as long as the version in `package.json` is the same).

So changes in the root folder were not respected by the example app, as it used a previously cached version of `jest-preset-angular`. This made the tests fail.

This is now fixed by not using cache in the `example` app. Using the cache and then just reinstalling `jest-preset-angular` somehow took way more time than just running the complete dependency installation.

---

For me everything looks good in this PR 👍 
Thanks again @sharikovvladislav for all the effort!

thymikee:
Gonna have a look at this tomorrow. I'm so impressed by the amount of work you all out into this!

AngularNoNgAttributesSnapshotSerializer.js
```diff
@@ -0,0 +1,34 @@
...
+
+const attributesToRemovePatterns = ['ng-reflect', '_nghost', '_ngcontent', 'ng-version'];
+
+const serialize = (node, ...rest) => {
+  const nodeCopy = node.cloneNode(true);
+  Object.values(nodeCopy.attributes)
+    .map(attribute => attribute.name)
+    .filter(attributeName =>
+      attributesToRemovePatterns
+        .some(removeAttributePattern => attributeName.startsWith(removeAttributePattern))
```
thymikee: how about extracting the filtering function out? it's the same as lines 23:25. We could also get rid of the extra `map` as well:

```js
const hasAttributesToRemove = attribute => 
  attributesToRemovePatterns.some(pattern => attribute.name.startsWith(pattern))
```

then:

```js
const serialize = (node, ...rest) => {
  // ...
    Object.values(nodeCopy.attributes)
      .filter(hasAttributesToRemove)
  // ...
}

const serializeTestFn = (val) => 
  val.attributes !== undefined && Object.values(val.attributes)
    .some(hasAttributesToRemove)

```

thymikee:
Also, please make sure CI is green (looks like something is leaking)

example/src/app/simple-with-styles/simple-with-styles.component.ts
```diff
@@ -0,0 +1,11 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-simple-with-styles',
+  templateUrl: './simple-with-styles.component.html',
+  // we have to setup styles this way, since simple styles/styleUrs properties will be removed (jest does not unit test styles)
+  ['styles']: [`
```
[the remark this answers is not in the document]
wtho: This is a workaround to avoid getting stripped the styles by the astTransformer.
If there are no styles to apply, Angular will not render `_ngcontent`, which we want to test here.

But a comment may be nice to explain this and not promote this syntax.

README.md
```diff
@@ -1,4 +1,4 @@
-# jest-preset-angular
+AngularSnapshotSerializer# jest-preset-angular
```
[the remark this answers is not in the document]
sharikovvladislav: Oops. Done.

README.md
```diff
@@ -69,6 +69,7 @@ module.exports = {
   },
   transformIgnorePatterns: ['node_modules/(?!@ngrx)'],
   snapshotSerializers: [
+    'jest-preset-angular/AngularNoNgAttributesSnapshotSerializer.js',
```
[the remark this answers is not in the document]
sharikovvladislav: Done

example/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "ng build",
     "test": "jest",
     "test:watch": "jest --watch",
-    "test:ci": "jest --runInBand",
+    "test:ci": "jest --runInBand --no-cache",
```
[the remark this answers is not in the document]
sharikovvladislav: @wtho looks like it was your change... do we still need it?

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.ts
```diff
@@ -0,0 +1,21 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-a-bit-more-complex',
```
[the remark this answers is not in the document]
sharikovvladislav: Oops. I already renamed that component to `medium` component. I forgot to commit deleted old files. Done.

example/src/app/simple-with-styles/simple-with-styles.component.ts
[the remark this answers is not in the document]
sharikovvladislav: > But a comment may be nice to explain this and not promote this syntax.

 I left comment there. Is it engough?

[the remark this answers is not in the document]
sharikovvladislav: Done

AngularNoNgAttributesSnapshotSerializer.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return jestDOMElementSerializer.serialize(nodeCopy, ...rest);
+};
+
+const serializeTestFn = (val) => {
+  return val.attributes !== undefined && Object.values(val.attributes)
+    .map(attribute => attribute.name)
+    .some(attributeName =>
+      attributesToRemovePatterns
+        .some(remotePattern => attributeName.startsWith(remotePattern))
```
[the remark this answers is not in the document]
sharikovvladislav: Done

AngularNoNgAttributesSnapshotSerializer.js
```diff
@@ -0,0 +1,34 @@
...
+        .some(removeAttributePattern => attributeName.startsWith(removeAttributePattern))
+    )
+    .forEach(attributeNameToRemove => nodeCopy.attributes.removeNamedItem(attributeNameToRemove));
+
+  return jestDOMElementSerializer.serialize(nodeCopy, ...rest);
+};
+
+const serializeTestFn = (val) => {
+  return val.attributes !== undefined && Object.values(val.attributes)
+    .map(attribute => attribute.name)
```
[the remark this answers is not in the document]
sharikovvladislav: I am not sure this will create some bottleneck about perfomance. Like we had `map` + `filter` and it is like `O(2n)`. One `filter` would be `O(n)`. May be I am not right but it is almost same for CPU. It not becomes `O(n^2)` which is really worse. But seems to me for people code is more readable using `map` + `filter`.

Done

AngularNoNgAttributesSnapshotSerializer.js
sharikovvladislav: Done

wtho:
I'll fix the CI later :+1:

wtho:
**How can I find out about leaks in the test runs?**

I ran the tests and they exceeded 10 minutes without output on [CircleCI](https://circleci.com/gh/wtho/jest-preset-angular/103).

Then I added the `--detectOpenHandles` flag and the tests ran [successfully](https://circleci.com/gh/wtho/jest-preset-angular/106).

The only change between the test runs was:
```diff
- yarn test:ci
+ yarn test:ci --detectOpenHandles
- yarn test:coverage
+ yarn test:coverage --detectOpenHandles
```

Unfortunately there is no information about possible leaks in the output.

example/src/app/simple-with-styles/simple-with-styles.component.ts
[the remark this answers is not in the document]
wtho: For me it's good. Any developer seeing this line will understand the reason behind this.

example/package.json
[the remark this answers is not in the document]
wtho: It can be removed!

README.md
```diff
@@ -84,7 +84,8 @@ module.exports = {
...
-- `"transformIgnorePatterns"` – unfortunately some modules (like @ngrx ) are released as TypeScript files, not pure JavaScript; in such cases we cannot ignore them (all node_modules are ignored by default), so they can be transformed through TS compiler like any other module in our project.
+- `"transformIgnorePatterns"` – unfortunately some modules (like @ngrx) are released as TypeScript files, not pure JavaScript; in such cases we cannot ignore them (all node_modules are ignored by default), so they can be transformed through TS compiler like any other module in our project.
+- `"snapshotSerializers"` - array of serializers which will be applied to snapshot the code. Note: by default angular adds some angular-specific attributes to the code (like `ng-reflect-*`, `ng-version="*"`, `_ngcontent-c*` etc). This package provides serializer to remove such attributes. This makes snapshots cleaner and more human-readable. To remove such specific attributes use `AngularNoNgAttributesSnapshotSerializer` serializer. You need to add `AngularNoNgAttributesSnapshotSerializer` serializer manually (like in `example` application in `package.json`).
```
thymikee: can we add a link to the exact lines in package.json?

thymikee:
Feel free to merge it @wtho :)

example/src/app/a-bit-more-complex-component/a-bit-more-complex.component.spec.ts
```diff
@@ -0,0 +1,34 @@
+/* tslint:disable:no-unused-variable */
```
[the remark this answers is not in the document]
wtho: Yeah!
But I would suggest to wait for Angular to make the official move to eslint, as this is still an Angular app example (might come in v8?).

Also see [the official tslint -> eslint issue](https://github.com/palantir/tslint/issues/4534)

ahnpnl:
Looking forward to test this in real projects 👍

sharikovvladislav:
@listepo-alterpost @barretodavid @isoden @iam4x @Lykathia @jfcere 

Any feedback

Lykathia:
I won't be able to check for a couple weeks now, unfortunately. Looking forward to it tho!

PR #99, opened 2017-12-11
Protect against potentially undefined nodes.

brianmcallister:
I was seeing some strange behavior popping up, where `node` was `undefined` for some of my components, which was errors being thrown about `hasOwnProperty` being called on `undefined`.

I don't really know a whole lot about the internals of Angular components, however, I did recently upgrade to Angular 5 so I suspect this might have something to do with that.

Inspecting the `val.componentRef._view.nodes` shows me that there are in fact multiple nodes in the array, and one is `undefined`. I tried simplifying my template down to a single `<div></div>`, but no luck.

All the tests are passing, but please let me know if there are any red flags around this sort of change.

### DIFF
--- a/AngularSnapshotSerializer.js
+++ b/AngularSnapshotSerializer.js
@@ -26,7 +26,7 @@ const print = (val, print, indent, opts, colors) => {
 
   const componentName = val.componentRef._elDef.element.name;
   const nodes = (val.componentRef._view.nodes || [])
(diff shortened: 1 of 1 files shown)


thymikee:
I'm ok with duck typing this, thanks for caring to send a PR!

thymikee:
Released 4.0.2 with this change, have fun 🙂

brianmcallister:
Awesome, thanks so much!

PR #101, opened 2017-12-16
Update dependencies

AhnpGit:
I updated dependencies for the main package.json and also updating to angular 5 for the example app


package.json
```diff
@@ -7,16 +7,17 @@
...
   "devDependencies": {
-    "jest": "^21.0.1",
-    "typescript": "^2.3.0"
+    "jest": "21.2.1",
+    "typescript": "~2.4.2"
   },
   "peerDependencies": {
-    "jest": "^21.0.1"
+    "jest": "^21.0.1",
+    "zone.js": "^0.8.18"
```
thymikee: Can you remove this? `jest-zone-patch` already has a peer dep on Zone.js

AhnpGit: sure will do. Thank you :)

package.json
```diff
@@ -7,16 +7,16 @@
   "author": "Michał Pierzchała <thymikee@gmail.com>",
   "license": "BSD-3",
   "dependencies": {
-    "jest-zone-patch": "^0.0.7",
-    "jsdom": "^11.3.0",
-    "ts-jest": "^21.0.0"
+    "jest-zone-patch": "^0.0.8",
+    "jsdom": "11.5.1",
+    "ts-jest": "21.2.4"
```
thymikee: btw, why are these deps locked?

package.json
```diff
@@ -7,16 +7,16 @@
...
   },
   "devDependencies": {
-    "jest": "^21.0.1",
-    "typescript": "^2.3.0"
+    "jest": "21.2.1",
```
thymikee: I prefer `^21.2.1`, we have a lockfile

package.json
```diff
@@ -7,16 +7,16 @@
...
+    "typescript": "~2.4.2"
```
thymikee: Why `~`? Is there something wrong with e.g `2.6`?

package.json
```diff
@@ -7,16 +7,16 @@
...
   },
   "peerDependencies": {
-    "jest": "^21.0.1"
+    "jest": "^21.2.1"
```
thymikee: This preset should still work with 21.0.1, so please revert this.

example/package.json
```diff
@@ -15,39 +15,39 @@
...
     "karma-jasmine": "~1.1.0",
     "karma-jasmine-html-reporter": "^0.2.2",
-    "protractor": "~5.1.0",
-    "ts-node": "~2.0.0",
-    "tslint": "~4.5.0",
-    "typescript": "~2.3.0"
+    "protractor": "~5.1.2",
+    "ts-node": "~3.2.0",
+    "tslint": "~5.7.0",
+    "typescript": "~2.4.2"
```
thymikee: These are the versions populated by `angular-cli`, right?

package.json
AhnpGit: At the moment Angular 5.0.x only supports up to 2.4.2, in the release 5.1.x they announced to support 2.5. Therefore I want to keep it consistent with Angular version supports. 5.0.x is stable now in my opinion so I'd like to keep TypeScript to be at the version supported by Angular 5.0.x. We can update to 2.5 later when Angular 5.1.x is stable.

package.json
AhnpGit: sure

package.json
AhnpGit: do you mean you prefer `^21.0.1` ?

package.json
thymikee: Cool, thanks for clearing this up!

example/package.json
AhnpGit: yes these are the versions coming from `angular-cli`. First I generated a new project with `angular-cli` and copied these versions from the new project's `package.json`

package.json
thymikee: Doesn't really matter, because it will resolve to 21.x.x. I'd go with `^21.2.1` if I had to choose :)

AhnpGit: then I'll keep this unchanged and only commit to change the version in `peerDependencies` back to `^21.0.1`

thymikee:
Does it pass tests locally for you? I get weird errors like:
```
The animation trigger "routerTransition" has failed to build due to the following errors:
     - The provided animation property "transform" is not a supported CSS property for animations
```
in `app.component.ts`

AhnpGit:
my bad, indeed it's the consequence after upgrading the example to Angular 5. Gonna check more to fix it

thymikee:
Thanks so much!

AhnpGit:
I added the workaround from example project of ngrx team. It works now but the CI build still fails, can you take a look because I don't know why it failed ?

thymikee:
Yay, you fixed it! :) Could you also apply this patch? It will make the example app actually compiling 😅 (it's a bug on current master)

```patch
From 5befea91d216094905516bcb6cd299cfd86f6d7d Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Micha=C5=82=20Pierzcha=C5=82a?= <thymikee@gmail.com>
Date: Sun, 17 Dec 2017 14:40:59 +0100
Subject: [PATCH] Fix example app not compiling

---
 example/src/app/app.component.ts | 4 +++-
 example/src/tsconfig.app.json    | 3 +--
 example/src/tsconfig.spec.json   | 4 ----
 3 files changed, 4 insertions(+), 7 deletions(-)

diff --git a/example/src/app/app.component.ts b/example/src/app/app.component.ts
index 63c47fb..a62d938 100644
--- a/example/src/app/app.component.ts
+++ b/example/src/app/app.component.ts
@@ -12,9 +12,11 @@ export class AppComponent {
   title = 'app works!';
   hasClass = true;
   variableWithPrecedingDolar = 1234;
+
+  ngOnInit() {}
 }
 
-function slideToLeft() {
+export function slideToLeft() {
   return trigger('routerTransition', [
     transition(':enter', [
       style({transform: 'translateX(200%)', position: 'fixed', width: '100%'}),
diff --git a/example/src/tsconfig.app.json b/example/src/tsconfig.app.json
index 5e2507d..4b3bfda 100644
--- a/example/src/tsconfig.app.json
+++ b/example/src/tsconfig.app.json
@@ -3,8 +3,7 @@
   "compilerOptions": {
     "outDir": "../out-tsc/app",
     "module": "es2015",
-    "baseUrl": "",
-    "types": []
+    "baseUrl": ""
   },
   "exclude": [
     "test.ts",
diff --git a/example/src/tsconfig.spec.json b/example/src/tsconfig.spec.json
index ca63ea9..cefee17 100644
--- a/example/src/tsconfig.spec.json
+++ b/example/src/tsconfig.spec.json
@@ -5,10 +5,6 @@
     "module": "commonjs",
     "target": "es5",
     "baseUrl": "",
-    "types": [
-      "jasmine",
-      "node"
-    ],
     "allowJs": true
   },
   "files": [
-- 
2.9.0

```

AhnpGit:
finished now :)

package.json
thymikee: Can we make `jsdom@^11.5.1` and `ts-jest@^21.2.4`?

AhnpGit: yes we can. Nothing breaks about this change. I did commit yarn.lock file for this change too

thymikee:
Thanks for sticking with me on this one!

PR #111, opened 2017-12-22
Add an example of testing a service

AhnpGit:
I made an example of testing a service. The service will get data from a backend API. In case it fails to get data, it will use console.error to print the error. If it succeeds, it will use console.log to print the log. I have a problem of making code coverage for the constructor of the service. Maybe you can help me with that. Thank you.

### DIFF
--- a/example/src/app/service/hero.service.spec.ts
+++ b/example/src/app/service/hero.service.spec.ts
@@ -1,0 +1,75 @@
+import { inject, TestBed } from '@angular/core/testing'
+import { HttpClient, HttpClientModule, HttpRequest } from '@angular/common/http'
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing'
+import { heroesUrl, HeroService } from './hero.service'
+
+describe('Service: GoogleBooks', () => {
+  let service: HeroService
+  let backend: HttpTestingController
+
+  const expectedData = {
+    id: 1,
+    name: 'Test hero'
+  }
+
+  beforeEach(() => {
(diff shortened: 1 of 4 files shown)


example/src/app/service/hero.service.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        HttpClientModule,
+        HttpClientTestingModule,
+      ],
+      providers: [
+        HeroService,
+      ],
+    })
+
+    spyOn(console, 'warn')
+    spyOn(console, 'error')
```
thymikee: can we just `jest.spyOn`? `global.spyOn` comes from Jasmine

example/src/app/service/hero.service.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  })
+
+  beforeEach(inject([ HttpClient, HttpTestingController ],
```
thymikee: Can't we get rid of inline injecting `HttpClient` and `HttpTestingController` here and in `afterEach()` and just move it to the first `beforeEach` call above?

example/src/app/service/hero.service.spec.ts
AhnpGit: oh ya we can replace that by jest.spyOn

example/src/app/service/hero.service.spec.ts
AhnpGit: ya, it looks cleaner indeed

thymikee:
Thank you so much!

Issue #125, opened 2018-02-20
Host bindings in the snapshot


yurii-sorokin:
Hi. Is there any plan to add attributes (classes, styles) added by `@HostBinding` to the snapshot?

thymikee:
No plans, but feel free to take a stab on it :)

yurii-sorokin:
Ok. I decided to create serializer as a separate package. It slightly different from serializer in this repo. It relies on debugNodes + existing react serializer. It's not stress tested yet, but you could take a look.

thymikee:
Share it :)

yurii-sorokin:
Oops, 🤦‍♂️.  The package on [github](https://github.com/yurii-sorokin/ngx-jest-snapshot-serializer) or on [npm](https://www.npmjs.com/package/ngx-jest-snapshot-serializer).

AhnpGit:
I like that you use serialize which is available since jest 21 👍

yurii-sorokin:
Also, as my serializer will break all existing tests (till the snapshot update of course), I have moved setup of serializers to jest config (PR #126). Now it would be possible to play with my serializer with this preset untouched.

yurii-sorokin:
I guess the issue could be closed.

PR #126, opened 2018-02-22
Move serializers setup to jest config to be possible to override them

yurii-sorokin:
#125

### DIFF
--- a/jest-preset.json
+++ b/jest-preset.json
@@ -24,5 +24,9 @@
   },
   "transformIgnorePatterns": [
     "node_modules/(?!@ngrx)"
+  ],
+  "snapshotSerializers": [
+    "<rootDir>/node_modules/jest-preset-angular/AngularSnapshotSerializer.js",
+    "<rootDir>/node_modules/jest-preset-angular/HTMLCommentSerializer.js"
   ]
 }
--- a/setupJest.js
+++ b/setupJest.js
@@ -18,6 +18,3 @@ getTestBed().initTestEnvironment(
   BrowserDynamicTestingModule,
   platformBrowserDynamicTesting()
 );
-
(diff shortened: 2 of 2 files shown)


jest-preset.json
```diff
@@ -24,5 +24,9 @@
   },
   "transformIgnorePatterns": [
     "node_modules/(?!@ngrx)"
+  ],
+  "snapshotSerializers": [
+    "<rootDir>/node_modules/jest-preset-angular/AngularSnapshotSerializer.js",
+    "<rootDir>/node_modules/jest-preset-angular/HTMLCommentSerializer.js"
```
thymikee: Shouldn't the order be different? Have you tested if that changes anything?

yurii-sorokin: Just tried both and seems there is no any difference. Will look how jest applies this config.

yurii-sorokin: I bet I've seen something about the order in jest docs, but can't find it now. Anyway, jest reverses the list:
```
config.snapshotSerializers
    .concat()
    .reverse()
    .forEach(path => {
      addSerializer(localRequire(path));
    });
```
So it's same as before.

AhnpGit:
nice 👍

thymikee:
published v5.1.0 with this change

ewantavenerdemo:
Does this mean I can now change the config for 'pretty-format'?

yurii-sorokin:
You can try [this serializer](https://github.com/yurii-sorokin/ngx-jest-snapshot-serializer) that uses the default serializers (built-in plugins for `pretty-format`). Basically, it transforms component fixture to react test object (mostly, as it does not convert dom nodes needed when using `[innerHTML]`). Which `jest-snapshot` understand well.

Or you can write your own serializer, for example, that uses `require('util').inspect` to serialize fixture or anything you like.

PR #131, opened 2018-03-07
Remove testRegex out of preset

AhnpGit:
Jest default set up testMatch to get all the test files to run. Removing testRegex out of the preset will bring flexibility to the users to set up their own testMatch/testRegex.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,8 @@ import './jestGlobalMocks'; // browser mocks globally available for every test
     "^.+\\.(ts|js|html)$": "<rootDir>/node_modules/jest-preset-angular/preprocessor.js"
   },
   "testMatch": [
-    "(/__tests__/.*|\\.(test|spec))\\.(ts|js)$"
+    "**/__tests__/**/*.+(ts|js)?(x)",
+    "**/?(*.)+(spec|test).+(ts|js)?(x)"
   ],
   "moduleFileExtensions": [
     "ts",
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ import './jestGlobalMocks'; // browser mocks globally available for every test
 * `<rootDir>` is a special syntax for root of your project (here by default it's project's root /)
 * we're using some `"globals"` to pass information about where our tsconfig.json file is that we'd like to be able to transform HTML files through ts-jest
 * `"transform"` – run every TS, JS, or HTML file through so called *preprocessor* (we'll get there); this lets Jest understand non-JS syntax
-* `"testMatch"` – we want to run Jest on files that matches this regex
+* `"testMatch"` – we want to run Jest on files that matches this glob
 * `"moduleFileExtensions"` – our modules are TypeScript and JavaScript files
 * `"moduleNameMapper"` – if you're using absolute imports here's how to tell Jest where to look for them; uses regex
 * `"setupTestFrameworkScriptFile"` – this is the heart of our config, in this file we'll setup and patch environment within tests are running
--- a/README.md
+++ b/README.md
@@ -45,6 +45,9 @@ import './jestGlobalMocks'; // browser mocks globally available for every test
   "transform": {
     "^.+\\.(ts|js|html)$": "<rootDir>/node_modules/jest-preset-angular/preprocessor.js"
   },
+  "testMatch": [
+    "(/__tests__/.*|\\.(test|spec))\\.(ts|js)$"
+  ],
   "moduleFileExtensions": [
     "ts",
     "js",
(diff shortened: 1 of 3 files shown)


jest-preset.json
```diff
@@ -8,7 +8,6 @@
   "transform": {
     "^.+\\.(ts|js|html)$": "<rootDir>/node_modules/jest-preset-angular/preprocessor.js"
   },
-  "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(ts|js)$",
   "moduleFileExtensions": [
```
thymikee: Ugh, but Jest will not match `*.ts` files by default. Why not changing it to `testMatch`?

AhnpGit: hmm, but if we add `testMatch`, people who want to use `testRegex` won't be able to run the tests. I think we should put `testMatch` to the README inside installation setup.

thymikee: But `testRegex` is already set and we write about it in README, so it will be the same situation. I'm not going to remove the default. One can always override it without any issues. Both cannot work with each other and Jest will warn you about it:

AhnpGit: ah ok then I'll adjust `testRegex` to be `testMatch` 👍

jest-preset.json
```diff
@@ -8,7 +8,9 @@
   "transform": {
     "^.+\\.(ts|js|html)$": "<rootDir>/node_modules/jest-preset-angular/preprocessor.js"
   },
-  "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(ts|js)$",
+  "testMatch": [
+    "(/__tests__/.*|\\.(test|spec))\\.(ts|js)$"
+  ],
```
thymikee: Can you adjust the readme and changelog as well?

README.md
```diff
@@ -45,7 +45,9 @@ import './jestGlobalMocks'; // browser mocks globally available for every test
   "transform": {
     "^.+\\.(ts|js|html)$": "<rootDir>/node_modules/jest-preset-angular/preprocessor.js"
   },
-  "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(ts|js)$",
+  "testMatch": [
+    "(/__tests__/.*|\\.(test|spec))\\.(ts|js)$"
```
thymikee: This is wrong. Here's the default one (where I just found a bug XD):
```
[
  '**/__tests__/**/*.js?(x)', 
  '**/?(*.)(spec|test).js?(x)' 
]
```
and we rather want to change it to:
```
[
  '**/__tests__/**/*.+(ts|js)?(x)', 
  '**/?(*.)+(spec|test).+(ts|js)?(x)' 
]
```

AhnpGit: oh those 2 are default from Jest 👍

AhnpGit:
I think the regex is not good enough. Just try `yarn pack` and install it into example app. It doesn't exclude `src/test.ts`

thymikee:
I think the second part should be 
```
"**/+(*.)(spec|test).(ts|js)?(x)"
```

AhnpGit:
- Tested with new project generated with CLI: OK
- Tested with example inside the repo: still fails with `src/test.ts`

thymikee:
See this: http://www.globtester.com/#p=eJzT0tLX1tDS09TWKC5ITa4pSS0u0dTT1igprskq1rTXqNAEAJdZCa0%3D&r=eJyzScksU0jOSSwutlVSsisuStYvSS0u0csqttEHytjZIEnnJpYkZ0DU5FbqoSgDAFtkGGY%3D&

AhnpGit:
Updated in the latest commit 👍

README.md
```diff
@@ -66,7 +69,7 @@ import './jestGlobalMocks'; // browser mocks globally available for every test
 * `<rootDir>` is a special syntax for root of your project (here by default it's project's root /)
 * we're using some `"globals"` to pass information about where our tsconfig.json file is that we'd like to be able to transform HTML files through ts-jest
 * `"transform"` – run every TS, JS, or HTML file through so called *preprocessor* (we'll get there); this lets Jest understand non-JS syntax
-* `"testRegex"` – we want to run Jest on files that matches this regex
+* `"testMatch"` – we want to run Jest on files that matches this regex
```
thymikee: this glob

jest-preset.json
```diff
@@ -8,7 +8,10 @@
   "transform": {
     "^.+\\.(ts|js|html)$": "<rootDir>/node_modules/jest-preset-angular/preprocessor.js"
   },
-  "testRegex": "(/__tests__/.*|\\.(test|spec))\\.(ts|js)$",
+  "testMatch": [
+    "**/__tests__/**/*.(ts|js)?(x)",
+    "**/+(*.)(spec|test).(ts|js)?(x)"
```
thymikee: we should also add `+`s before the `(..|..)` groups, because that's a valid glob. What we use currently at Jest (and I filed a PR to fix it: https://github.com/facebook/jest/pull/5744) uses an extra feature from `micromatch` library.
```
"**/__tests__/**/*.+(ts|js)?(x)",
"**/+(*.)+(spec|test).+(ts|js)?(x)"
```

thymikee:
Thanks for sticking with me!

AhnpGit:
Thank you too. I learnt something new about glob as well

thymikee:
Me too!

Issue #96, opened 2017-11-23
Zone already loaded error in version ^4.0.1 and ^5.x.x


jbgraug:
Whenever there is a dependency on zone.js. Jest terminates.
It does not happen on v3.0.1. 

To reproduce just place
import 'zone.js' 
in setup-jest.ts

thymikee:
Zone is already imported and lives in global scope (when you're using `setupTestFrameworkScriptFile` hook, it's not there yet in `setupFiles` which executes earlier), it's forbidden to redeclare it, so don't do it manually.

jbgraug:
I did it manually just to show that it breaks. My project uses a lib. Wich
uses zone.js and whenever I import a file which has any direct or indirect
dependence on it, it breaks.

thymikee:
Can you tell show something more about the error?

jbgraug:
If i drill down into the lib using zone.js, it imports zone.js like this ``` require("zone.js");  ``` 
which is different from ``` require('zone.js/dist/zone.js'); ``` the one you are using and then it crashes.
Any test having a dependency with zone.js raises the error below, even though the app builds/run properly.
Test suite failed to run
    Zone already loaded.
      at node_modules/zone.js/dist/zone-node.js:32:15
      at node_modules/zone.js/dist/zone-node.js:646:3
      at Object.<anonymous>.FUNCTION (node_modules/zone.js/dist/zone-node.js:9:65)
      at Object.<anonymous> (node_modules/zone.js/dist/zone-node.js:12:2)
      ...
```

Hotell:
is this still an issue ? not experiencing this at all -> angular 5.x

thymikee:
@jbgraug can we close it?

jbgraug:
Still got the same error, my solution was to stick to version 3.x
Tried version 5, still failing with the same error.

jbgraug:
What's the problem of importing zone.js the "right way"?
In zone.js package.json they already have both properties "main" and "browser" set up.
Why override this functionality in importing "zone.js/dist/zone" when the right version is going to automatically be picked up?

Please take a look here https://github.com/angular/zone.js/issues/963

patrickmichalina:
This break testing with newer versions of angularfire2

thymikee:
Does it unbreak with #135? I didn't have the capacity to test it yet

patrickmichalina:
[...]

PR #135, opened 2018-03-29
Revert /dist/zone.js import to fix issues/96

jbgraug:
This fixes the error that happens when you have 3rd. party libraries also using zone.js (for both node and browser). 

https://github.com/thymikee/jest-preset-angular/issues/96.

### DIFF
--- a/setupJest.js
+++ b/setupJest.js
@@ -2,7 +2,7 @@
 
 require('core-js/es6/reflect');
 require('core-js/es7/reflect');
-require('zone.js/dist/zone.js');
+require('zone.js');
 require('zone.js/dist/proxy.js');
 require('zone.js/dist/sync-test');
 require('zone.js/dist/async-test');


thymikee:
I'll yet need to check that locally if it doesn't break tests. We had some issues with this import and one test was breaking.

Skysplit:
@jbgraug @thymikee `require('zone.js/dist/zone.js');` casues memory leaks while running jest.
Tested using simple `expect(true).toBe(true)` and each time the test suite was run, it increased memory heap size by about 10 megabytes (sometimes even more).

Replacing it to simple `require('zone.js')` fixed this issue.

anjmao:
I also noticed memory leak on zone.js, but after changing import `require('zone.js');` all tests which are using fakeAsync and tick are not working anymore.

anjmao:
Here in node-zone.js https://github.com/angular/zone.js/blob/master/lib/node/node.ts#L31 timers are not fully patched because isMix returns true. 

```ts
// we are in electron of nw, so we are both browser and nodejs
// Make sure to access `process` through `_global` so that WebPack does not accidentally browserify
// this code.
export const isMix: boolean = typeof _global.process !== 'undefined' &&
    {}.toString.call(_global.process) === '[object process]' && !isWebWorker &&
    !!(isWindowExists && internalWindow['HTMLElement']);
``` 

As a workaround I was able to fix it by temporary removing HTMLElement from jsdom window.
```ts
const htmlEl = window['HTMLElement'];
window['HTMLElement'] = undefined;
require('zone.js/dist/zone-node.js');
window['HTMLElement'] = htmlEl;
```
Now at least memory is not growing crazy and gc is collecting it sometimes (still not perfect) and fakeAsync is working correctly.

AhnpGit:
@anjmao I saw Angular repo they setup Jasmine with node-zone.js https://github.com/angular/angular/blob/master/tools/cjs-jasmine/index-tools.ts and in https://github.com/angular/angular/blob/master/tools/testing/init_node_spec.ts
Can those links provide some hints regarding to this PR ?

AhnpGit:
@anjmao I found this [jasmine fix async](https://github.com/angular/zone.js/blob/1ba851989ccb6907df49bba37ee24ab60adf13a9/lib/jasmine/jasmine.ts#L71) in `zone.js`. I think we need something similar like that

dauledk:
@thymikee thanks for the quick answer. I friend recommended this preset with angular at a meetup, on the promise it would be easy to configure. It was (thanks to this awesome repo), but is now breaking because of zone-js. I am really liking Jest, but the whole objective from my part is running tests faster, without having to configure anything myself 🤓

thymikee:
Yea I get this is a pain, but there's no clear solution for now. Hopefully we'll come up with something

dauledk:
Ok awesome. I would also like to help to solve the problem, but maybe it is not the perfect solution for a first contribution :)

PR #139, opened 2018-04-08
Fix `STYLE_URLS_REGEX` to not break on multiple new lines

### LINKED ISSUE #136: Multi-lined `styleUrls` seems to break tests.
Components that are declared like this:

```
@Component({
    selector: 'whatever',
    styleUrls: [
        '../some-styles.scss',
        '../some-more-styles.scss',
        '../even-more-styles.scss',
    ]
```

break the regex of the preprocessor, resulting in:

```
SyntaxError: The string did not match the expected pattern.
      
      at XMLHttpRequest.open [as __zone_symbol__open] (../../node_modules/jsdom/lib/jsdom/living/xmlhttprequest.js:470:15)
```

Putting the `styleUrls` all on one line fixes this issue.

Seems to me that the regex could just be `styleUrls:\s*\[[^\]]*\]`? Unless I'm really missing something. Let me know and I will make a PR.

phenomnomnominal:
Fixes #136

### DIFF
--- a/__tests__/preprocessor.test.js
+++ b/__tests__/preprocessor.test.js
@@ -26,6 +26,25 @@ const sources = [
     '../media-box.component.scss',
     './media-box-h0.component.scss'
   ],
+})`,
+  `@Component({
+  selector: 'xc-media-box-h0',
+  templateUrl: 'media-box-h0.component.html',
+  styleUrls: [
+    '../../box.component.scss',
+    '../media-box.component.scss',
+    './media-box-h0.component.scss'
+  ],
+})`,
+  `@Component({
+  selector: 'xc-media-box-h0',
+  templateUrl: 'media-box-h0.component.html',
+  styleUrls: [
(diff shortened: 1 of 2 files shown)


__tests__/preprocessor.test.js
```diff
@@ -26,6 +26,25 @@ const sources = [
...
+})`,
+  `@Component({
+  selector: 'xc-media-box-h0',
+  templateUrl: 'media-box-h0.component.html',
+  styleUrls: [
+    '../../../box.component.scss',
+    '../../box.component.scss',
+    '../media-box.component.scss',
+    './media-box-h0.component.scss'
+  ],
```
thymikee: Looks like this case is not necessary. Can you remove?

phenomnomnominal: Can if you want - it's intention was to make sure I wasn't just making it work for 3 external stylesheets.

PR #173, opened 2018-07-22
Upgrade ts-jest, @types/jest

ahnpnl:
- Upgrade `ts-jest` to 23.0.1
- Upgrade `@types/jest` to 23.3.0
- Upgrade devDependencies for `jest`

I noticed that relative path resolving for `tsconfig` seems to be broken by `ts-jest`. I'm not sure if we should wait for them to fix that issue before we can release 6.0

### DIFF
--- a/__tests__/tsconfig.spec.json
+++ b/__tests__/tsconfig.spec.json
@@ -1,14 +1,10 @@
 {
-  "extends": "./example/tsconfig.json",
+  "extends": "../example/tsconfig.json",
   "compilerOptions": {
     "outDir": "../out-tsc/spec",
     "module": "commonjs",
     "target": "es5",
     "baseUrl": "",
     "allowJs": true
-  },
-  "include": [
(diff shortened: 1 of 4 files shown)


thymikee:
Maybe they changed something, is this intended or a regression?

ahnpnl:
I think it's more about a regression. Here is the [reported issue](https://github.com/kulshekhar/ts-jest/issues/625). 

I did try `jest` 23 with our current preset version and it seems to work for my company projects at least. Not sure if other people encounter something differently. From my point of view, we still can wait for `ts-jest` to fix that issue first.

thymikee:
Cool, let's wait then

ahnpnl:
@thymikee done upgrading to latest `ts-jest` major version. I think it's stable now on their side.

thymikee:
Thank you @ahnpnl, I'll trust you on that :) Releasing v6.0.0 in a minute

PR #182, opened 2018-08-08
support backtick quoted templateUrl

SEIAROTg:
As explained in readme, the current approach to processing `templateUrl` is a scrappy regex. However, the current regex only supports single quote and double quote.

This PR add backtick support to it.

### DIFF
--- a/__tests__/preprocessor.test.js
+++ b/__tests__/preprocessor.test.js
@@ -55,7 +55,19 @@ const sources = [
     '../media-box.component.scss',
     './media-box-h0.component.scss'
   ],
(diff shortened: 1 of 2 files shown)


ahnpnl:
Hi, thank you for supporting us. Would you please also add the test case for template string and double quotes to preprocessor.spec.js ? That would be great.

SEIAROTg:
Sure. Added.

thymikee:
@ahnpnl feel free to merge it :)

ahnpnl:
👍

PR #183, opened 2018-08-08
Fix README.md to include moduleNameMapper

### LINKED ISSUE #181: Tests won't run with angular/cli@6.1.2
Hello,

When creating a new app via angular cli, and then following the instructions on adding jest-preset-angular, the tests will not run.
My pair and I have been trying to figure out what the problem is for about a week, but we have had no luck other than reverting the version of angular cli.

To reproduce: 
1. `npm install -g @angular/cli` (get the latest)
1. `ng new my-app && cd my-app`
1. `npm install --save-dev jest jest-preset-angular`
1. Create the `setupJest.ts` and `jestGlobalMocks.ts` files, provided by `jest-preset-angular`
1. Add jest configurations for `preset`, `setupTestFrameworkFile` and `testURL` to `package.json`
1. `npx jest`
1.  Test fail :(
```
Test suite failed to run

    /Users/pivotal/workspace/last-hope/src/setupJest.ts:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import 'jest-preset-angular';
                                                                                                    ^^^^^^^^^^^^^^^^^^^^^

    SyntaxError: Unexpected string

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:403:17)
```

This appears to be complaining about having been translated to the wrong ECMAScript version (i.e., es2015 vs. commonjs).  Here is the generated `tsconfig.json`:

```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "baseUrl": "./",
    "outDir": "./dist/out-tsc",
    "sourceMap": true,
    "declaration": false,
    "module": "es2015",
    "moduleResolution": "node",
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "target": "es5",
    "typeRoots": [
      "node_modules/@types"
    ],
    "lib": [
      "es2017",
      "dom"
    ]
  }
}
```

We think the issue is that jest cannot handle the es2015, but the problem still exists if we change `module` to `commonjs`.

If we rollback the version of angular cli to 6.0.8, we get no issues.
No one else seems to have encountered this, but we can reproduce it reliably.

comatose-turtle:
Without the moduleNameMapper config option, jest will not run on the current version of angular CLI (you will get a `SyntaxError: Unexpected string`).  We discovered that this is necessary through testing the various differences between a newly generated angular CLI project and the `/example` project in this repo.

Fixes #181

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,10 @@ import './jestGlobalMocks'; // browser mocks globally available for every test
 {
   "jest": {
     "preset": "jest-preset-angular",
-    "setupTestFrameworkScriptFile": "<rootDir>/src/setupJest.ts"
+    "setupTestFrameworkScriptFile": "<rootDir>/src/setupJest.ts",
+    "moduleNameMapper": {
+      "@lib/(.*)": "<rootDir>/src/lib/$1"
+    }
   }
 }
 ```


ahnpnl:
Actually that is the configuration between jest and typescript. This configuration has been mentioned in README of ts-jest. I'd prefer editing our README to point back to ts-jest README for this issue.

Issue #110, opened 2017-12-21
Add more tests to example app


thymikee:
In the `example` dir of this repository we have couple of tests to check if we didn't break some common testing cases. It is a very simple app and I treat it as sanity check before releasing. Also, a good reference for anybody wanting to see how to write specific tests. It would be nice to extend this test suite, to cover more cases, so we have more confidence in further releases. 

Since I'm not actively involved in any Angular project for quite some time and I still want this project to be alive, I reach out to you as a community to help with this task. I'm sure we can make this project more stable with every release.

If anyone is interested in contributing tests, please let me know by either commenting in this thread or sending a PR 🙂

ahmedifhaam:
I need some help I am currently using JEST to test an angular 9 app, There i want to mock service classes which are dependencies for my component. I want my tests to take a default mock from __mocks__ folder and if i need i will override  with mockImplementaion(). currently its working if i import the mock file but i think the  way jest documentation says it should pick the mock when i call jest.mock('filepath'). In my case its not working

wtho:
@ahmedifhaam best would be you either create a PR with the test case not working in example app, or you create another new issue with your specific request and provide a minimal reproduction in form of an own, linked repository, that will enable us to help you quickly.

rfprod:
- https://github.com/thymikee/jest-preset-angular/pull/1182
- https://github.com/thymikee/jest-preset-angular/pull/1183
- https://github.com/thymikee/jest-preset-angular/pull/1181
- https://github.com/thymikee/jest-preset-angular/pull/1185
- https://github.com/thymikee/jest-preset-angular/pull/1184

ahnpnl:
Closes via all mentioned MRs above.

wtho:
Just wondering - does the V10 example app also get the tour of heroes tests?
Totally ok for me if not, but just wanted to make sure this was not just left out my mistake, but on purpose.

ahnpnl:
Angular 10 support will be dropped on 24th this month so we don’t need to add tests to example app Angular 10 :)

PR #188, opened 2018-08-26
Add HeroesComponent example with mocked service provider (#110)

ParkerM:
Related to Issue #110
An example of how to mock a service provider's return values using `jest.fn()` and rxjs' `Subject` without having to bother with mocking low-level HTTP responses. 
I placed the component in its own `heroes` folder which is kind of awkward since the Hero model and service are in the more generic `service` folder, but I didn't want to take too many liberties in moving stuff around.  

Other small changes:
* Removed `'rxjs'` from the import-blacklist as it was causing linter warnings on existing code.
* Fixed incorrect describe name on existing HeroService spec.

### DIFF
--- a/example/src/app/app.module.ts
+++ b/example/src/app/app.module.ts
@@ -7,13 +7,15 @@ import { AppComponent } from './app.component';
 import { CalcComponent } from './calc/calc.component';
 import { SimpleComponent } from './simple/simple.component';
 import { OnPushComponent } from './on-push/on-push.component';
+import { HeroesComponent } from './heroes/heroes.component';
 
 @NgModule({
   declarations: [
     AppComponent,
     CalcComponent,
     SimpleComponent,
-    OnPushComponent
+    OnPushComponent,
+    HeroesComponent
   ],
   imports: [
     BrowserModule,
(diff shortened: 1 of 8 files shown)


thymikee:
Honestly I'm against this kind of testing. You mock the service, so whenever you break it, test won't catch it 🤷‍♂️. I usually advice to mock request library/API itself, because that's what your code doesn't have control over – you can however mock request to return anything and check whether your service reacts properly.

ParkerM:
Yeah for sure, this type of mocking can be dangerous if the service itself is not also tested. This is under the assumption that the service itself is also properly tested (i.e. via HTTP mocks).  

Ideally the component should not know or care where the service is getting its data from. By mocking the low-level HTTP responses in the component spec the tests become tightly coupled with dependencies that the component itself does not know exist. This can be especially problematic for services that are used in many places. For example, a small change to an external API could require updates to every spec with the service in its dependency chain. However, if the contract between the app and the external API is solely owned by the respective service then only that single spec would need to be updated and the developers can go fast forever.

Another advantage of this style (especially with angular stuff) is that other distant dependencies can be omitted from the spec, which can minimize the required resources and speed up run time.

thymikee:
Yea, I'm a fan of testing integrations not units :D But I don't see a reason we shouldn't merge that, it's just an example :)

ahmedifhaam:
Can you update this same example to work by automocking a service class ?

Issue #195, opened 2018-09-19
Not compatible with ts-jest@23.10.0


FrozenPandaz:
## Current Behavior

`ts-jest@23.10.0` (published today) is a rewrite of the package and has many breaking changes

```
yarn list v1.9.4
warning Filtering by arguments is deprecated. Please use the pattern option instead.
├─ jest-preset-angular@6.0.0
│  └─ ts-jest@23.10.0
```

## Expected Behavior

Short term, lock `ts-jest` to `~23.1.0`.

Long term, ensure compatibility with `ts-jest@^23.10.0`.

ahnpnl:
PR is welcome 🙂

byronmejia:
Thanks for the fix @FrozenPandaz ! Such a shame that  [`kulshekhar/ts-jest`](https://github.com/kulshekhar/ts-jest) can't follow semver...

How can we help with ensuring compatibility with 23.1.0? Any places to start at, or any tips?

ahnpnl:
Hi @byronmejia

I contacted with ts-jest's contributors and got some clues to start but I am on vacation now so I don't have much time to look into it. You can contact @huafu to discuss a point to start :)

byronmejia:
Awesome! Cheers. :)

huafu:
@danil-z and others, here is my WIP PR #201, since I have no time for now to continue working on it. The big stuff remaining is explained in the header of the preprocessor.js file which itself needs to be removed.

PR #196, opened 2018-09-19
fix: restrict version of ts-jest to ~21.1.3

FrozenPandaz:
fixes short term https://github.com/thymikee/jest-preset-angular/issues/195

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "dependencies": {
     "@types/jest": "^23.3.1",
     "jest-zone-patch": "^0.0.8",
-    "ts-jest": "^23.1.3"
+    "ts-jest": "~23.1.3"
   },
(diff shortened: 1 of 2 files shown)


ahnpnl:
23.1.3* 😁

thymikee:
Can't we use `~23.9.0`?

ahnpnl:
There is sadly no 23.9.. they jumped from 23.1 to 23.10

thymikee:
Oh, that's unexpected :D

FrozenPandaz:
Oops.. Sorry about that, I edited the message right as you published :(

FrozenPandaz:
My guess is that it is some sort of indication they rewrote the whole package... 

See https://github.com/kulshekhar/ts-jest#versioning

PR #200, opened 2018-10-16
Update jest-preset.json

schoes:
add new config for ts jest to support new versions

### DIFF
--- a/jest-preset.json
+++ b/jest-preset.json
@@ -1,9 +1,9 @@
 {
   "globals": {
     "ts-jest": {
-      "tsConfigFile": "src/tsconfig.spec.json"
-    },
-    "__TRANSFORM_HTML__": true
+      "tsConfig": "src/tsconfig.spec.json",
+       "stringifyContentPathRegex": true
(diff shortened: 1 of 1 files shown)


jest-preset.json
```diff
@@ -1,9 +1,9 @@
 {
   "globals": {
     "ts-jest": {
-      "tsConfigFile": "src/tsconfig.spec.json"
-    },
-    "__TRANSFORM_HTML__": true
+      "tsConfig": "src/tsconfig.spec.json",
+       "stringifyContentPathRegex": true
+    }
```
thymikee: we should also bump our dependency of ts-jest to be newer, right? as I assume this is breaking

ahnpnl: Indeed it is more like breaking change. I think the whole changes should include changing jest-preset, bumping ts-jest, adjust preprocessor. I'm not sure if changing jest-preset is enough.

schoes: right. sorry will do some more work to get it done right

PR #201, opened 2018-10-17
WIP to make it compatible with ts-jest 23.10 and remove regexp replacers

huafu:
DO NOT MERGE, this is a WIP which I do not have time to continue just right now

see #195


thymikee:
Awesome, waiting for your availability then :)

wtho:
## So to push this PR a bit
(and maybe start helping out)

As I see what is still missing, it is just a single transformer has to look for the `@Component`-Decorator and checks if there are
* `styles`
* `styleUrls`
* `template`
* `templateUrl`

Then the transformer has to
* Replace the styles with an empty string
* Replace styleUrls with an empty array
* Add the inline template somehow (did this work in the old processor?)
* Add the HTML file as require statement to the node

I wonder if adding the HTML as `require` is sufficient to make ts-jest load the html, or if it has to be added to the context in some other way, any ideas?

ahnpnl:
inline template works in the current preprocessor. Regarding to adding the HTML file I have no idea because the current preprocessor doesn't do that as far as I'm concerned.

wtho:
Ok, as @huafu stated in the [comments of the PR's preprocessor.js](https://github.com/thymikee/jest-preset-angular/blob/2aea6d12d713ecb898b203c8ee8cd98de71e9871/preprocessor.js), the inline html part is handled by ts-jest.

As far as I understand it, the current preprocessor checks for `templateUrl` and replaces it with an inlined `template: require(...)`-statement:
https://github.com/thymikee/jest-preset-angular/blob/abfb339cbe1e4e76bc68de66c9cfadc7d5da3588/preprocessor.js#L2
https://github.com/thymikee/jest-preset-angular/blob/abfb339cbe1e4e76bc68de66c9cfadc7d5da3588/preprocessor.js#L11

I will have a try to implement it as an AST transformer.

ahnpnl:
hi @wtho, thank you very much for your effort. In my opinion, I think:
- Fork the current work of @huafu and create a new PR and refer this PR from it seems to be a good way to do.
- Indeed the changes will be breaking because of `ts-jest`. Since Angular already released v7, we can release your changes in v7 of this preset as well. Would you please add this change to `CHANGELOG` ?
- I think it's ok to leave a `.ts` file and compile it into `.js`.
- Indeed the example app refers to npm package. I think the ideal situation we can have is running pre-build script before running the all the tests inside this repo. But I think it should be in another PR. We can run some tests against your branch to make sure the changes are fine before merging into master.

PR #204, opened 2018-11-28
Implement astTransformer - Compatibility with ts-jest 23.10

### LINKED ISSUE #195: Not compatible with ts-jest@23.10.0

wtho:
Closes #195 and closes #201

### Breaking Change
This is a breaking change as Configuration Interface changes ([see ts-jest v23.10 docs](https://kulshekhar.github.io/ts-jest/user/config/#ts-jest-options))

This PR is built upon @huafu's #201, kudos to him for setting up everything and pointing out what was still to be done. Transformer and test are also heavily inspired by his examples in ts-jest. Also see the PR for more information.

### Implementation Details
Like the Regex, all the transformer does is transform
* `templateUrl: <PATH>` to `template: require(<PATH>)`
* `styles: <ANYTHING>` to `styles: []` 
* `styleUrls <ANYTHING>` to `styleUrls: []`

For more information see the comments in the [transformer file](https://github.com/wtho/jest-preset-angular/blob/55bcfdb993d1dc511611553995ff566d64ea6944/src/InlineHtmlStripStylesTransformer.ts) or #201.

### Edge Cases
I had to make some decisions on how strict the astTransformer is, I decided for this, but I am open to discussion:
**Works**
```ts
import { Component as Cmp } from '@angular/core';
@Cmp({
  templateUrl: './some-path.html'
})
class AngularComp { }
```
Caveat - this also gets transformed
```ts
@SomeOtherDecorator({
  templateUrl: './some-path.html'
})
class SomeClass { }
```

**Does not work**
```ts
const componentArgs = {
  templateUrl: './some-path.html'
}
@Component(componentArgs)
class AngularComp { }
```

### Additional Notes
* Run unit tests for the transformer with `npm test`
* The transformer was written in TypeScript. A `prepare`-script for was added, which compiles it to JS before publishing and after `npm install`.
* `preprocessor.js` and its unit tests were removed.


wtho:
To make the example-app's tests pass again, I would suggest to remove their execution from the circleci tasks for now and add them after this version got published the first time.

After that the tests will run again referencing `jest-preset-angular` from `package.json` avoiding version conflicts and Angular can be bumped to v7.

Note that the `jest-preset-angular`-reference in `example/package.json` was changed (which does not work this way either), but changing it back to the package in `node_modules` will also not fix the issue for now.

ahnpnl:
hi @wtho, I tried to check out your branch, ran `npm run build` and then `npm pack` then install the `.tgz` file to example app. Then I ran test for the components inside example app, I observed that `simple.component.ts` test failed with this error

```
yarn run v1.12.3
$ jest
 FAIL  src/app/simple/simple.component.spec.ts
  ● Test suite failed to run

    Cannot find module 'simple.component.html' from 'simple.component.ts'

      3 | @Component({
      4 |   selector: 'app-simple',
    > 5 |   templateUrl: 'simple.component.html',
        |   ^
      6 |   styleUrls: ['./simple.component.css']
      7 | })
      8 | export class SimpleComponent implements OnInit {

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:221:17)
      at src/app/simple/simple.component.ts:5:3
      at Object.<anonymous> (src/app/simple/simple.component.ts:8:1)
```

The same error I saw when running this `.tgz` file against my company's project. Not all components' tests failed but only a few of them.

README.md
```diff
@@ -40,29 +43,30 @@ import './jestGlobalMocks'; // browser mocks globally available for every test
...
   },
   "testMatch": [
-    "**/__tests__/**/*.+(ts|js)?(x)",
-    "**/+(*.)+(spec|test).+(ts|js)?(x)"
-  ],
-  "moduleFileExtensions": [
-    "ts",
-    "js",
-    "html"
+    "**/__tests__/**/*.ts?(x)",
```
ahnpnl: why these 2 regexes were modified ? The old regexes works

ahnpnl:
@thymikee are you familiar with ast transformation ? I don't have much experience with it.

README.md
wtho: As much as I can see there is no reason to change them. This was done by **huafu** in the commit before, and I do not know why, maybe js-files are not supported by ts-jest anymore (which I doubt), but I have not tested it.

Will undo it in the next commit.

wtho:
You are correct with the test! I did not know paths like `templateUrl: 'page.html'` were allowed in angular and have to be transformed before putting them to `require`.

The transformer now adds `./` if the path does not start with `./`, `../` or `/`. The simple-component test should work now.

The `prepare` Script is a special one that will automatically run after you `npm install` without any additional args and before `npm publish`, so it is not required to call `npm run build`.

To make the example-app tests pass you have to run:
```bash
cd example
npm install

# install version of this PR in example-app, will also get ts-jest@23.10
# and build InlineHtmlStripStylesTransformer.js from ts
# (did not know about that, but works! npm ftw!)
npm install wtho/jest-preset-angular#ts-jest-23-10

npm test
```

I wanted to add this workaround to the CI-Tests as well until next publication, but yarn does not build the package after installed from github directly as npm does, so it would have been more complicated.

I has issues with jest's caching, as I ran the tests long ago, so I run the tests with `npm test -- --no-cache` to make sure the transformer gets called every time.

Also one more test in the example app fails for me, seems to be related to the Serializers (on Linux, Node v11):
**AppComponent > snaps**:  
  ```
  -       aaa \$1234
  +       aaa $1234

  -      \`test'chars""
  +      `test'chars""
  ```

index.js
```diff
@@ -1 +1,20 @@
...
+require('core-js/es6/reflect');
+require('core-js/es7/reflect');
+require('zone.js/dist/zone.js');
+require('zone.js/dist/proxy.js');
+require('zone.js/dist/sync-test');
+require('zone.js/dist/async-test');
+require('zone.js/dist/fake-async-test');
+require('jest-zone-patch');
+// const AngularSnapshotSerializer = require('./AngularSnapshotSerializer');
+// const HTMLCommentSerializer = require('./HTMLCommentSerializer');
```
thymikee: remove dead code

circle.yml
```diff
@@ -16,8 +16,8 @@ dependencies:
     - ~/.cache/yarn
   override:
     - yarn install --no-progress
-    - cd example && yarn install --no-progress
+#    - cd example && yarn install --no-progress
 test:
   override:
     - yarn run test:ci
-    - cd example && yarn run test:ci && yarn run test:coverage
+#    - cd example && yarn run test:ci && yarn run test:coverage
```
thymikee: example tests need to run, please uncomment

__tests__/InlineHtmlStripStylesTransformer.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    const out = transpile(CODE_WITH_STYLE_URLS)
+
+    const compiledBeginningOfComponent = 'var AngularComponent'
+
+    const output = out.outputText
+    expect(output).toContain(compiledBeginningOfComponent)
+
+    // just compare class part, generated decorator function is not interesting
+    const classPart = output.substr(output.indexOf(compiledBeginningOfComponent))
+    expect(classPart).toEqual(
```
thymikee: I think it would make sense to snapshot the output of `classPart` and provide a descriptive name of what's expected to be inside this output

__tests__/InlineHtmlStripStylesTransformer.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    'body: { display: none }',
+    'html: { background-color: red }'
+  ]
+})
+export class AngularComponent {
+}
+`
+
+const CODE_WITH_ALL_DECORATOR_PROPERTIES = `
+import { Component } from '@angular/core';
```
thymikee: some indentation wouldn't hurt (and will be easy to adjust once output tests are converted to snapshots

jest-preset.js
```diff
@@ -0,0 +1,29 @@
...
+    'ts-jest': {
+      tsConfig: '<rootDir>/src/tsconfig.spec.json',
+      stringifyContentPathRegex: '\\.html$',
+      astTransformers:[
+        require.resolve('./InlineHtmlStripStylesTransformer')
+      ]
+    },
+  },
+  transform: {
+    '\\.(ts|html)$': 'ts-jest',
```
thymikee: we don't provide transformer for JS files now, is it safe to assume any JS code in 3rd party dependencies is valid ES2015 code?

.gitignore
```diff
@@ -1 +1,2 @@
 node_modules
+InlineHtmlStripStylesTransformer.js
```
thymikee: would be cool to convert the source code to TS some day and have a `build/` directory where we could point the package. not in scope of this PR of course.

__tests__/InlineHtmlStripStylesTransformer.test.ts
```diff
@@ -0,0 +1,214 @@
+/*
+ * Code is inspired by
+ * https://github.com/kulshekhar/ts-jest/blob/25e1c63dd3797793b0f46fa52fdee580b46f66ae/src/transformers/hoist-jest.spec.ts
+ * original license:
```
thymikee: We already include this license within our package because we use `ts-jest` as our dependency. Please remove it here and in `InlineHtmlStripStylesTransformer.ts`. The "code inspiration" notice is good to stay of course.

src/InlineHtmlStripStylesTransformer.ts
```diff
@@ -0,0 +1,260 @@
...
+        case TEMPLATE_URL:
+          // we can reuse the right-hand-side literal from the assignment
+          let templatePathLiteral = assignment.initializer
+
+          // fix templatePathLiteral if it was a non-relative path
+          if (isStringLiteral(assignment.initializer)) {
+            const templatePathStringLiteral: StringLiteral = assignment.initializer;
+            // match if it starts with ./ or ../ or /
+            if (templatePathStringLiteral.text &&
+              !templatePathStringLiteral.text.match(/^(\.\/|\.\.\/|\/)/)) {
```
ahnpnl: I haven't tested this improvement yet but by only reading the codes, it's not so generic to me. Because it's not so sure where the users can put their template, it might be ./ , ../../ or some other places within the project. I think it's better to resolve the path within the project or at least within src/app

wtho: True!
I just checked the old regex mechanism, it would also not support `templateUrl: 'src/page.html'` and transform it to `./src/page.html`.
When the astTransformer hits, it only gets `config-set` injected. Luckily, the ConfigSet object seems to have a method [resolvePath](https://github.com/kulshekhar/ts-jest/blob/984eb8558887c727a6be6eaa02825cb4b882b1bb/src/config/config-set.ts#L700), I will see later if this serves our purpose on checking if `require` will be able to find the file later on.

__tests__/InlineHtmlStripStylesTransformer.test.ts
wtho: :+1: will update

.gitignore
wtho: So I keep the generated `InlineHtmlStripStylesTransformer.js` in `.gitignore` or do you suggest I should commit it as well?
The js-files could be easily moved to the `src`-folder, tsc will just keep them compiled. But would add a lot of noise to this PR, which is quite big already. I would prefer to do it another time.

circle.yml
wtho: As discussed in [#204](https://github.com/thymikee/jest-preset-angular/pull/201#issuecomment-442205239) and [described earlier](https://github.com/thymikee/jest-preset-angular/pull/204#issuecomment-442272185), the tests will not run until this code itself is published as `jest-preset-angular`, as it is self referencing but going through a breaking change right now.

I can only make them run using
```
cd example && \
npm install && \
npm install wtho/jest-preset-angular#ts-jest-23-10 && \
yarn run test:coverage
```
or
```
cd example && \
yarn install && \
yarn remove jest-preset-angular#ts-jest-23-10 && \
yarn add wtho/jest-preset-angular#ts-jest-23-10 && \
cp ../InlineHtmlStripStylesTransformer.js node_modules/jest-preset-angular && \
yarn run test:coverage
```
as the example-app references the version on npm, which is in version 6, but this is a breaking change.

Maybe you have another idea?

jest-preset.js
wtho: Fixed this one, was taken over from previous commit this PR is built on top of.

circle.yml
thymikee: We used to link the library, but it failed in some cases if I remember well. Should revisit that, either by using yarn link or installing the preset from directory instead of npm package

src/InlineHtmlStripStylesTransformer.ts
wtho: Ok I looked into the issue a bit and saw the ts-compiler itself (which is also available in the transformer) offers many more path-tools than expected:
`toPath, normalizePath, normalizePathAndParts, getDirectoryPath, isRootedDiskPath, isDiskPathRoot, convertToRelativePath, getPathComponents, reducePathComponents, getNormalizedPathComponents, getNormalizedAbsolutePath, getPathFromPathComponents, getRelativePathFromFile, getRelativePathFromDirectory, getRelativePathToDirectoryOrUrl, ensurePathIsNonModuleName, combinePaths, resolvePath, comparePathsCaseSensitive, comparePathsCaseInsensitive, comparePaths, containsPath, extensionFromPath, tryGetExtensionFromPath, getAnyExtensionFromPath, fullTripleSlashReferencePathRegEx, fullTripleSlashAMDReferencePathRegEx, getExternalModuleNameFromPath, getOwnEmitOutputFilePath, getDeclarationEmitOutputFilePath, getSourceFilePathInNewDir, getOutputPathsFor, updateMissingFilePathsWatch, getDefaultLibFilePath`

But it feel like it would be an own task to figure out which methods help us to figure out if the file path is relative to the transformed file or relative to the project root using any path-definition inside `tsconfig.json`.

I quickly checked `resolvePath`, but that always appends the file to the project root.

Would you be ok with opening a new issue and for now keep the same behavior as the regex did before?

ahnpnl: Of course opening another PR for it is always fine by us👌 The only problem is we will need to create a temporary branch for this PR to merge in instead of master. Because master branch at the moment doesn't encounter this issue therefore we want to keep the state like it is right now. Until we solve the issue with this resolve path we can merge the branch of this PR into master. Do you agree @thymikee ?

circle.yml
wtho: Linking alone did not work for me, some packages were still missing, but I found a workaround.

### Did not work
On linking or setting `"preset": "/my/absolut/path/of/jest-preset-angular"` in package.json:
```
    Cannot find module 'zone.js/dist/zone.js' from 'index.js'

      2 | require('core-js/es6/reflect');
      3 | require('core-js/es7/reflect');
    > 4 | require('zone.js/dist/zone.js');
        | ^
```
After adding zone.js to the root-project it complains about `@angular/core`, then `rxjs`, `@angular/platform-browser-dynamic`, `@angular/platform-browser`, `@angular/common`, `@angular/compiler`, ...

On `"preset": ".."` in package.json:
```
● Validation Error:
  Module jest-preset-angular should have "jest-preset.js" or "jest-preset.json" file at the root.
```

### Worked
From the root of this repo, after both example app and root had run their initial installs:
```
npm install example/node_modules/@angular/common example/node_modules/@angular/compiler example/node_modules/@angular/core example/node_modules/@angular/platform-browser example/node_modules/rxjs example/node_modules/zone.js

npm link
cd example
npm link jest-preset-angular
npm test
```
I like this approach as it makes sure it uses the versions of the example app when running the tests and we do not pollute the root package.json (ok, when running tests we do, but generally not)

If you are ok with this, I will update the CI configuration and test it on the circleci.

__tests__/InlineHtmlStripStylesTransformer.test.ts
wtho: :+1: for using snapshots.

I would remove the substringing and have the decorator-bootstrapping in the snapshot as well, it does not distract as much in there.

wtho:
Thank you for your feedback @thymikee @ahnpnl, feel free to check out the latest changes!

circle.yml
thymikee: Just merged it, feel free to rebase and get over this issue :)

src/InlineHtmlStripStylesTransformer.ts
ahnpnl: hmm ok. I hope we make the users happy with this breaking change before they find out this hidden bug :)

package.json
```diff
@@ -7,20 +7,32 @@
   "author": "Michał Pierzchała <thymikee@gmail.com>",
   "license": "MIT",
   "dependencies": {
-    "@types/jest": "^23.3.1",
+    "@types/jest": "^23.3.5",
```
ahnpnl: since this one is in this PR, would you mind put this to `23.x` and `ts-jest` to `23.10.5` ? I know `23.x` works but `ts-jest` `23.10.x` I'm not sure it works

circle.yml
wtho: Rebased, and works, awesome!

src/InlineHtmlStripStylesTransformer.ts
wtho: #### Quick Intro AST and the transformer
(Ignore if you know this already)

If you play with a [ast-viewer](https://ts-ast-viewer.com/) by putting in a simple Angular Component you can get familiar with the AST that TSC will parse form the code. It is a big tree of nested nodes.

Every AST node will have a `kind`-property, which is a SyntaxKind (all of them listed in [typescript.d.ts](https://github.com/Microsoft/TypeScript/blob/8ddb2b61d2c996ab442b282ab9b36195ed697c13/lib/typescript.d.ts#L78)).

For every SyntaxKind there is a `is<KIND>(node)`-method, e. g. `isStringLiteral(node)`. I use these in the transformer, [e. g. here](https://github.com/thymikee/jest-preset-angular/pull/204/files#diff-d8c58610f25910c24023a249bf950abbR137), and can then safely assume the node is of that kind and has the corresponding fields, so there is no need to check against the number 'manually'.

Once called `getMutableClone()` on a node it is possible to modify inner nodes. Nodes can also be programmatically created, as explained [here in the TS docs](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#creating-and-printing-a-typescript-ast). We have to do this in the transformer to e. g. [create the `require()`-call](https://github.com/thymikee/jest-preset-angular/pull/204/files#diff-d8c58610f25910c24023a249bf950abbR179).

Being familiar with this enables you to write custom TSLint rules as well.

ahnpnl:
thanks 👍

wtho:
> Strange CI still failed because one of the snapshots

Yeah it's odd, I want to figure this out before merging.

wtho:
Ok so this is also a breaking change of ts-jest 23.10, of which they are probably not aware of
**cc @huafu**

Previously ts-jest used to handle html just calling `JSON.stringify(htmlFileContent)` in [preprocess](https://github.com/kulshekhar/ts-jest/blob/240649bb14a1eb45c5f0454bd56f8315ec2262e2/src/preprocess.ts#L30)

now they create a module using ```result = `module.exports=${JSON.stringify(source)}` ``` in [ts-jest-transformer](https://github.com/kulshekhar/ts-jest/blob/a6bcec48ef28791235b3eba0b3f2bd1a944ba5b8/src/ts-jest-transformer.ts#L118) and process it like a file using babel.

This changes the output in the snapshots as (I guess) babel removes the escaping.

ahnpnl:
It comes from [stringifyContentPathRegex](https://kulshekhar.github.io/ts-jest/user/config/stringifyContentPathRegex) which to keep backward compat of the old option `__HTML_TRANSFORM__`. Even though the option itself is to keep the existing test working but potentially breaks some snapshot files.

ahnpnl:
- Breaking changes should be shortly explained in CHANGLOG and link to this PR. Perhaps in this case of snapshot you can refer it to ts-jest documentation. 
- I saw the comment from @thymikee about transpile file and convert file in this project to ts file. I guess it's ok to leave it in .gitignore.
- The 3rd issue about the test of app component I am not aware of that error. We have a lot of rooms to improve tests in example app.

thymikee:
>Breaking changes should be shortly explained in CHANGLOG and link to this PR. Perhaps in this case of snapshot you can refer it to ts-jest documentation.

Yup, plus if there's a migration path, feel free to also add it to the changelog :)

wtho:
I updated the changelog and added a migration guide (ts-jest does apparently not offer one, so I explained the changes to be made in the config).

Btw the snapshots does not differ because of `JSON.stringify`, but because of the regex `ESCAPE_TEMPLATE_REGEX` added in #34 is not used anymore. It is not required anymore, as the transformer works on the AST instead of on character-based regexes. I mentioned this in the changelog as well.

Feel free to edit the changelog in any way before the v7 release.

package.json
```diff
@@ -9,12 +9,11 @@
   "dependencies": {
     "@types/jest": "^23.3.10",
     "jest-zone-patch": "^0.0.8",
-    "ts-jest": ">=23.10 <23.11"
+    "ts-jest": "~23.10.0"
   },
   "devDependencies": {
-    "@types/node": "^10.12.10",
```
thymikee: I think yarn.lock should get updated as well?

wtho: Yeah it should :smile:

wtho:
> @wtho let me know if you'd like to help more on the project and become a collaborator just like @ahnpnl who also does tremendous job keeping this project alive.

Yeah, I would love to! Also thank you both for helping me through this PR so far.
It is great you are so quick with responses and giving constructive feedback, this is how open source is fun!

thymikee:
Published this as `7.0.0-alpha.1`. If anyone is interested in testing this, feel free to install `jest-preset-angular@next` :)

Issue #170, opened 2018-07-10
Cannot read property trim of undefined


Nxt3:
This happens if I don't import `NoopAnimationsModule` into the spec file.

```
ERROR { TypeError: Cannot read property 'trim' of undefined
          at applyKeyframeAnimation (/Users/nxt3/code/my_app/node_modules/@angular/animations/bundles/animations-browser.umd.js:4063:46)
          at ElementAnimationStyleHandler.Object.<anonymous>.ElementAnimationStyleHandler.apply (/Users/nxt3/code/my_app/node_modules/@angular/animations/bundles/animations-browser.umd.js:4022:9)
          at CssKeyframesPlayer.Object.<anonymous>.CssKeyframesPlayer.init (/Users/nxt3/code/my_app/node_modules/@angular/animations/bundles/animations-browser.umd.js:4193:22)
          at CssKeyframesPlayer.Object.<anonymous>.CssKeyframesPlayer.play (/Users/nxt3/code/my_app/node_modules/@angular/animations/bundles/animations-browser.umd.js:4199:14)
          at TransitionAnimationPlayer.Object.<anonymous>.TransitionAnimationPlayer.play (/Users/nxt3/code/my_app/node_modules/@angular/animations/bundles/animations-browser.umd.js:3710:91)
          at /Users/nxt3/code/my_app/node_modules/@angular/animations/bundles/animations-browser.umd.js:3490:20
          at Array.forEach (<anonymous>)
```

How do I globally mock this so I don't have to import `NoopAnimationsModule` every time?

This is the culprit function:

```
function applyKeyframeAnimation(element, value) {
    var anim = getAnimationStyle(element, '').trim(); //right here
    var index = 0;
    if (anim.length) {
        index = countChars(anim, ',') + 1;
        value = anim + ", " + value;
    }
    setAnimationStyle(element, '', value);
    return index;
}
```

devakone:
Running into the same issue. Isn't it best practices to use the `NoopAnimationsModule`?

jsonberry:
Same issue here, and unfortunately importing NoopAnimationsModule isn't doing the trick. 

No repo available to upload for this one.

The error:
```
    TypeError: Cannot read property 'trim' of undefined

      at applyKeyframeAnimation (../../packages/animations/browser/src/render/css_keyframes/element_animation_style_handler.ts:79:46)
[...]

PR #206, opened 2018-11-28
fix: add mock for BrowserAnimationsModule

splincode:
Closed: https://github.com/thymikee/jest-preset-angular/issues/170

### DIFF
--- a/example/src/jestGlobalMocks.ts
+++ b/example/src/jestGlobalMocks.ts
@@ -35,3 +35,27 @@ Object.defineProperty(document.body.style, 'transform', {
     };
   },
 });
+
+
+/**
+ * ISSUE: https://github.com/thymikee/jest-preset-angular/issues/170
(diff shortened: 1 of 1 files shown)


thymikee:
Btw, can we add a small smoke test?

ahnpnl:
> Btw, can we add a small smoke test?

Agree 👍

thymikee:
@wtho is this still needed provided with jsdom v13 or v14?

wtho:
It is already included in jsdom [since vs 11.12](https://github.com/jsdom/jsdom/blob/master/Changelog.md), good catch!
(it was added in cssstyle v1.0, which was udpated in jsdom v11.12).

I also checked on version 13 and the properties area already set.

Yeah, we can update jsdom on occasion, but there are no game-changing features/bugfixes in v14 as far as I can judge.

thymikee:
Awesome! I'll close this PR then. Thanks @splincode!

PR #207, opened 2018-12-02
chore: use local preset from example app

thymikee:
This change makes the example properly test the local preset instead of published version.

cc @wtho


wtho:
:+1:

PR #212, opened 2018-12-06
Migrate CI config to CircleCI v2

wtho:
CircleCI v1 is no longer supported and will stop working at all after August

Read more at [circleci.com/sunset1-0](https://circleci.com/sunset1-0/)

This PR contains the new config. It runs inside a node docker image instead of a fully fledged EC2 instance - so it is faster to setup and has yarn preinstalled.


.circleci/config.yml
```diff
@@ -0,0 +1,40 @@
...
+            - ~/.yarn
+            - ~/.cache/yarn
+            - ./node_modules
+            - ./example/node_modules
+      - run:
+          name: Test Library
+          command: |
+            yarn test:ci
+      - run:
+          name: Test Example App
```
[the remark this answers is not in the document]
wtho: Yeah, tried first put it in one step, but I preferred the immediate overview in the circleci-panel, in which part of the app the tests would fail. See [this example](https://circleci.com/gh/wtho/jest-preset-angular/10)

If you prefer them to be summarized, I can reduce them to one step.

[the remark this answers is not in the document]
wtho: Oh now I got you, overlooked **workflows**
Will set that up :smile:

thymikee:
I've pointed it to `release/7` branch for now, maybe it will kick in once we merge into a branch on this repo.

.circleci/config.yml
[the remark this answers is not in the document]
wtho: Ok, so I made it work.
The overview is more extendable and tests run parallel, but currentlly we lose about 30s for not running in one workspace (workspace has to be persisted and attached between different jobs).

I still prefer it, as it gives a better overview, especially if we might later choose to add linting, generating docs, ... (see [workflow example](https://circleci.com/workflow-run/f90a4a1e-1317-42d3-9be8-38b8967d7feb) from react-native-paper)

.circleci/config.yml
```diff
@@ -0,0 +1,66 @@
...
+            # the install afterwards will only install packages with different versions
+            - yarn-cache-{{ .Branch }}-{{ checksum "yarn.lock" }}
+            - yarn-cache-{{ .Branch }}-
+      - run:
+          name: Install dependencies
+          command: |
+            yarn --version
+            yarn install --frozen-lockfile
+            cd example
+            yarn install --frozen-lockfile
```
thymikee: I think install-dependencies should only include the core library, example can be installed in `unit-tests-example` step, what do you think?

.circleci/config.yml
```diff
@@ -0,0 +1,66 @@
+# Note: all paths are relative to the default workspace ~/project
+version: 2.1
+executors:
+  node-11:
```
thymikee: can we rename to node-10, since that's what it is?

wtho: Set executor to node v11

.circleci/config.yml
wtho: :+1:

.circleci/config.yml
thymikee: Please use stable releases, 10 is good.

PR #224, opened 2019-01-26
feat: Jest 24

### LINKED ISSUE #223: upgrade peer dependency to jest 24
Jest 24 was released yesterday, Jan 25. It looks like there are quite a few breaking changes. What needs to be done to help this library upgrade it's pear dependency to Jest 24?

thymikee:
This is a PoC. Jest 24 works without any changes from our side. It added some defaults to TS but we use slightly different ones (e.g. no `.tsx`).

However, until `ts-jest` updates, we'll likely need to hold off with the upgrade, because of peer dep warnings.

Fixes #223

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 InlineHtmlStripStylesTransformer.js
+*.log
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ import './jestGlobalMocks'; // browser mocks globally available for every test
 {
   "jest": {
     "preset": "jest-preset-angular",
-    "setupTestFrameworkScriptFile": "<rootDir>/src/setupJest.ts"
+    "setupFilesAfterEnv": ["<rootDir>/src/setupJest.ts"]
   }
 }
 ```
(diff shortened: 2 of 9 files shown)


xxxtonixxx:
it's necessary to add [it/test.todo](https://jestjs.io/blog/2019/01/25/jest-24-refreshing-polished-typescript-friendly#testtodo) support to fix this error: `TypeError: it.todo is not a function` :+1:

ahnpnl:
> it's necessary to add it/test.todo support to fix this error: TypeError: it.todo is not a function

is it to add to `zone-patch` @thymikee ?

thymikee:
Yup, it's there. I'll update the PR later today (or feel free to do that for me instead, you can push to this branch)

FrozenPandaz:
Hi, it seems like development on this has been paused. Can this be prioritized because of the security vulnerability with `jest@23.x.x`? https://www.npmjs.com/advisories/786.

Are there any blockers?

ahnpnl:
I don't see any blockers. Only one feature nice to have is it/test.todo support

ahnpnl:
Just finished reading all and it was merged 🤣

thymikee:
Released as `7.0.0-alpha.5`, please test it and maybe it's time to release a stable already? ;)

ahnpnl:
Will try immediately 👌

FrozenPandaz:
> maybe it's time to release a stable already? ;)

Looks good to me, was able to integrate into Nx. Looking forward to the stable :)

ahnpnl:
apart from performance downgraded, I don't see any issues.

thymikee:
We discuss perf regression here https://github.com/facebook/jest/pull/8032

If you have anything valuable to add, that would be great!

ahnpnl:
I think originally the performance was affected by `ts-jest` already when we switch to ast transformer, jest 24 just adds more on top of it...

thymikee:
Just to be sure, you're on 24.1, right? We had a caching regression in 24.0 fixed in minor

ahnpnl:
ya I was on 24.1. In my opinion mainly the performance is caused by ast transformer.

FrozenPandaz:
Is a stable version blocked by these performance issues?

thymikee:
Nope, I think we should finally release it :p

wtho:
👍  from my side

ahnpnl:
Actually the performance is not so bad. Only the 1st time jest needs to recache things and after that 2nd time all the tests finished with a more less similar amount of time for my work's project comparing to v6

PR #214, opened 2018-12-13
Import jest zone patch

JLHwung:
This PR is addressed to https://github.com/thymikee/jest-zone-patch/pull/9#issuecomment-445772763.

For the sake of complete git history, I merge the commits into this repository.

There is another LICENSE issue: The `jest-zone-patch` is licensed under BSD-3-Clause while this repository under MIT. I keep the original BSD LICENSE on this PR and left it as-is.

PR #258, opened 2019-04-07
📝 Improve docs

krokofant:
* jest-zone-patch is embeded (#214) as a dependency and should not be mentioned as being installed at the same time as this package is not found in package.json, package-lock.json or yarn.lock - and can be presumed missing.
* @types/jest might not be installed so install it explicitly
* Specify how to avoid karma conflicts (ref #233)
  - Otherwise consumers might get `TypeError: Cannot read property 'prototype' of undefined` while jest processes new angular cli projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@ _Note: This preset does not suport AngularJS (1.x). If you want to set up Jest w
 
 ```bash
 yarn add -D jest jest-preset-angular
+# or
+npm install -D jest jest-preset-angular
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,6 +41,10 @@ _Note: feel free to copy the `jestGlobalMocks.ts` file from the example director
 }
 ```
 
+### Avoid karma conflicts
```
thymikee:
```suggestion
### Avoid karma conflicts

```

PR #261, opened 2019-04-11
feat: transform 'styles' only in decorator

### LINKED ISSUE #254: InlineHtmlStripStylesTransformer is not strict enough
## Current Behavior

_Any_ object literals have their property assignments transformed thus causing source code such as:

```ts
console.log({
  styles: ['styles.css']
})
```

to print out

```ts
{
  styles: []
}
```

## Expected Behavior

Only object literals passed to `@angular/core`'s `@Component` decorator should have it's property assignments transformed.

wtho:
This PR is another implementation change in the transformation behavior. It is a proposal and open for discussion.

### Motivation

See #254 

Until now we transformed 'templateUrl', 'styleUrls' and 'styles'
everywhere they were assigned.

This even transformed code like this:
`console.log({ styles: {...}})` => `console.log({ styles: [] })`

The new implementation of the AstTransformer only transforms property
assignments to 'styles' if it is placed inside a decorator.

`templateUrl` and `styleUrls` are still transformed everywhere if assigned to a string, as Angular might expect them to be an actual Angular-Template or CSS code. This will crash the tests, if they are passed to the `@Component` decorator.

### Example
A new unit test explains the new implementation:

Untransformed:
```ts
  const assignmentsToNotBeTransformed = {
    styles: [{
      color: 'red'
    }]
  };

  const assignmentsToBeTransformed = {
    styleUrls: ['./some-styles.css'],
    templateUrl: './some-styles.css'
  };
```
Transformed:
```js
var assignmentsToNotBeTransformed = {
    styles: [{
            color: 'red'
        }]
};
var assignmentsToBeTransformed = {
    styleUrls: [],
    template: require('./some-styles.css')
};
```

### Caveats
There were always caveats and there still will be.
* This preset replaces styles wherever it decides they are not required. This design-decision was made due to the unit-test nature of jest. To test your styles, better use an e2e framework. The style stripping should speed up your tests, but might hinder you from testing applied styles.
* `templateUrl` and `styleUrls` are transformed anywhere in the code. This enables you to declare them somewhere in your unit tests and then plug them into a component, e. g. using  
  ```js
    TestBed.configureTestingModule({
      declarations: [
        AlertFollowComponent,
      ],
    }).overrideComponent(AlertFollowComponent, {
      set: {
        templateUrl: '../__mocks__/alert-follow-stub.component.html',
      },
    });
  ```
  At the same time, there will be side-effects on variables named `templateUrl` and `styleUrls` if you use them in a different context.

**EDIT**:
* The implementation lead to introduce more code in the ASTTransformer, this means, while it is more granular in its behavior, it is also more difficult to maintain.

Closes #254

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 node_modules
-InlineHtmlStripStylesTransformer.js
+InlineFilesTransformer.js
+StripStylesTransformer.js
+TransformUtils.js
 *.log
 .idea
--- a/__tests__/StripStylesTransformer.test.ts
+++ b/__tests__/StripStylesTransformer.test.ts
@@ -1,0 +1,61 @@
+/*
+ * Code is inspired by
+ * https://github.com/kulshekhar/ts-jest/blob/25e1c63dd3797793b0f46fa52fdee580b46f66ae/src/transformers/hoist-jest.spec.ts
+ *
+ */
+
+import * as tsc from 'typescript';
+import * as transformer from '../StripStylesTransformer';
+
+const CODE_WITH_STYLES_AND_OTHER_ASSIGNMENTS = `
+  import { Component } from '@angular/core';
+
+  @SomeDecorator({
+    value: 'test',
+    styles: [
+      ':host { background-color: red }'
+    ],
+  })
+  @Component({
+    templateUrl: './page.html',
+    styleUrls: [
+      './fancy-styles.css',
+      './basic-styles.scss'
+    ],
+    styles: [
+      'body { display: none }',
+      'html { background-color: red }'
+    ],
+    unaffectedProperty: 'whatever'
+  })
+  export class AngularComponent {
+  }
+`;
+
+const CODE_WITH_ASSIGNMENT_OUTSIDE_DECORATOR = `
+  const assignmentsToNotBeTransformed = {
+    styles: [{
+      color: 'red'
+    }]
+  };
+`;
+
+const createFactory = () => {
+  return transformer.factory({ compilerModule: tsc } as any);
+};
+const transpile = (source: string) =>
+  tsc.transpileModule(source, { transformers: { before: [createFactory()] } });
+
+describe('inlining template and stripping styles', () => {
+  it('should not strip styleUrls assignment', () => {
+    const out = transpile(CODE_WITH_STYLES_AND_OTHER_ASSIGNMENTS);
+
+    expect(out.outputText).toMatchSnapshot();
+  });
+
+  it('should not transform styles outside decorator', () => {
+    const out = transpile(CODE_WITH_ASSIGNMENT_OUTSIDE_DECORATOR);
+
+    expect(out.outputText).toMatchSnapshot();
+  });
+});
--- a/__tests__/__snapshots__/InlineFilesTransformer.test.ts.snap
+++ b/__tests__/__snapshots__/InlineFilesTransformer.test.ts.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`inlining template and stripping styles should handle all decorator assignments in differently named decorators 1`] = `
+exports[`inlining template and stripping styleUrls should handle all decorator assignments in differently named decorators 1`] = `
 "\\"use strict\\";
 var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
     var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
@@ -24,7 +24,7 @@ exports.AngularComponent = AngularComponent;
 "
 `;
 
-exports[`inlining template and stripping styles should handle all transformable decorator assignments 1`] = `
+exports[`inlining template and stripping styleUrls should handle all transformable decorator assignments 1`] = `
 "\\"use strict\\";
 var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
     var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
@@ -44,7 +44,10 @@ var AngularComponent = /** @class */ (function () {
         core_1.Component({
             template: require('./page.html'),
             styleUrls: [],
-            styles: [],
+            styles: [
+                'body { display: none }',
+                'html { background-color: red }'
+            ],
             unaffectedProperty: 'whatever'
         }),
         SomeOtherDecorator({
@@ -57,7 +60,7 @@ exports.AngularComponent = AngularComponent;
 "
 `;
 
-exports[`inlining template and stripping styles should handle templateUrl in test file outside decorator 1`] = `
+exports[`inlining template and stripping styleUrls should handle templateUrl in test file outside decorator 1`] = `
 "\\"use strict\\";
 Object.defineProperty(exports, \\"__esModule\\", { value: true });
 var testing_1 = require(\\"@angular/core/testing\\");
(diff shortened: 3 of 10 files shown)


wtho:
Actually it might be smart to split them up into two transformers, didn't think about it earlier

ahnpnl:
I think we should do benchmarking with this change as well.

wtho:
Yeah!

I still definitely want to split it up before merging.

wtho:
* Split the AstTransformers in two and implemented the more careful behavior for `styles`, it only gets removed in `@Component`-Decorators.
* Assignments with the names `templateUrl` and `styleUrls` will still be transformed anywhere.

Will fix the `flatMap` issue in the next days.

@ahnpnl do you have any suggestions how to approach the benchmarking?

wtho:
* replaced `flatMap` with `reduce` in transformer
* added README and CHANGELOG notes
* added tests for `styles`
* rebased on top of master

src/StripStylesTransformer.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (!ts.isClassDeclaration(node) || !node.decorators) {
+      return [];
+    }
+
+    return node.decorators
+      .map(dec => dec.expression)
+      .filter(ts.isCallExpression)
+      .filter(callExpr => ts.isIdentifier(callExpr.expression) && callExpr.expression.getText() === COMPONENT)
+      .reduce((acc, nxtCallExpr) => acc.concat(
+        ...nxtCallExpr.arguments
```
ahnpnl: I don't know if slice is faster than spread operator when running with node.

src/StripStylesTransformer.ts
```diff
@@ -0,0 +1,150 @@
...
+          .filter(ts.isObjectLiteralExpression)
+          .reduce((acc, nxtArg) => acc.concat(
+            ...nxtArg.properties
+              .filter(ts.isPropertyAssignment)
+              .filter(propAss => ts.isIdentifier(propAss.name))
+              .filter(propAss =>
```
ahnpnl: Can we combine these 2 filter conditions ?

src/StripStylesTransformer.ts
```diff
@@ -0,0 +1,150 @@
...
+    const mutableNode = ts.getMutableClone(node)
+    const assignments = getInDecoratorPropertyAssignmentsToTransform(mutableNode)
+
+     assignments.forEach(assignment => {
+      switch ((assignment.name as Identifier).text) {
+        case STYLES:
+          // replace initializer array with empty array
+          assignment.initializer = ts.createArrayLiteral()
+          break
+      }
```
ahnpnl: I don't think break will stop forEach, the loop still continues unless that is the intention ?

.gitignore
```diff
@@ -1,4 +1,6 @@
 node_modules
-InlineHtmlStripStylesTransformer.js
+InlineFilesTransformer.js
+StripStylesTransformer.js
+TransformUtils.js
```
thymikee: I think it's time for a `build/` directory – ideally in a separate PR :)

__tests__/StripStylesTransformer.test.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+const transpile = (source: string) =>
+  tsc.transpileModule(source, { transformers: { before: [createFactory()] } });
+
+describe('inlining template and stripping styles', () => {
+  it('should have correct signature', () => {
+    expect(transformer.name).toBe('angular-component-strip-styles');
+    expect(typeof transformer.version).toBe('number');
+    expect(transformer.version).toBeGreaterThan(0);
+    expect(typeof transformer.factory).toBe('function');
```
thymikee: Isn't TS guarantee these typeof checks? I think they're redundant (here and there)

__tests__/__snapshots__/InlineFilesTransformer.test.ts.snap
```diff
@@ -44,7 +44,10 @@ var AngularComponent = /** @class */ (function () {
         core_1.Component({
             template: require('./page.html'),
             styleUrls: [],
-            styles: [],
+            styles: [
+                'body: { display: none }',
+                'html: { background-color: red }'
```
thymikee: colons after `body` and `html` are not expected

src/StripStylesTransformer.ts
thymikee: Instead a `.find` could be used to return early

src/StripStylesTransformer.ts
```diff
@@ -0,0 +1,150 @@
...
+   * Traverses the AST down inside a decorator to a styles assignment
+   * returns it in an array.
+   */
+  function getInDecoratorPropertyAssignmentsToTransform(node: Node) {
+    if (!ts.isClassDeclaration(node) || !node.decorators) {
+      return [];
+    }
+
+    return node.decorators
+      .map(dec => dec.expression)
```
thymikee: How big this array can be realistically? If it's 1000+ elements, it may make sense to squash multiple filters/maps into a single call. The best thing would be to actually measure on a cold cache :)

.gitignore
wtho: Will start with that once this is done!

__tests__/StripStylesTransformer.test.ts
wtho: Yeah, I agree! Removed.

src/StripStylesTransformer.ts
wtho: Actually there were more cases before in the `SWITCH`-statement, the `forEach` should still iterate over all the elements.

I replaced it with a simple if-statement.

src/StripStylesTransformer.ts
wtho: I did use `Array.prototype.concat(acc, nxtCallExpr.arguments...)` to avoid the spread operator.

src/StripStylesTransformer.ts
wtho: **TL;DR**
It usually has only 1-4 elements, this is in my opinion the shortest and most readable version.

**Explanation**
`node.decorators` is the array that contains all decorators of a class.
Most classes anyway have only one decorator, in that case the whole function will be traversed only once, maybe even aborted, if it is a wrong decorator or the expressions inside are wrong.
In the two `flatMaps` (here `reduce`) they map to all the property assignments in the decorator, so in each decorator there are like 2-4 I would say. These get filtered out in the very end to only `styleUrls` and `templateUrl`, so the last two filters are called for each property assignment in a `@Component` decorator in the codebase.

What I personally value more is that else might have to refactor the code in some months and it is better to understand it easier and to write less code.

Finally TypeScript has better type support when using a single statement in a filter (e. g. `.filter(ts.isCallExpression)`), it tells me without cast that the argument in the next `map`/`filter` is of the type `CallExpression`, due to the return type of the function `is CallExpression`. The same does not happen when chaining with `&&`, so otherwise there will be more casting and assignments to casted variables.

Hope this explains my view, but I'm open to make changes.

**Benchmarking**
I have also no data to support this, but I believe the performance-wise expensive part of the transformers are when things are modified. I think this is much more expensive than iterating over some parts of the asts.

With this new implementation we modify less `styles` assignments and so this should be faster. But I am also sure that the tasks done by `ts-jest`, like building the initial ast and transpiling the code afterwards are much more heavier than these small transformations. Probably the optimizations of the v8-engine weigh much more than code changes in the transformers.

I am critical about benchmarking in general, I think to get a meaningful result we should test a bigger codebase with more components. Maybe there is a big open source Angular project out there?

In the CI-test-process we could clone a repo and test it with disabled jest-cache.

src/StripStylesTransformer.ts
wtho: Combined the latter ones.

src/StripStylesTransformer.ts
thymikee: That makes perfect sense, thanks for the detailed writeup :)

wtho:
In last force-push (to `351e919`), after @thymikee's approval I fixed minor issues (that I thought I pushed before):
* fixed CSS statements like  `body: { display: none }` - removed `:`
* removed tests that test basic signatures (what TS does anyway)
* joined two filter assignments in `StripStylesTransformer.ts`
* replaced `switch` with single `case` by `if`

thymikee:
@Ahnpnl anything to add or can e merge it? :)

ahnpnl:
LGTM 🎉

PR #269, opened 2019-05-17
Update changelog for v7.1.0

ahnpnl:
I totally forgot we released new serializer in 7.1.0

cc @thymikee


thymikee:
@ahnpnl @wtho can you both ping me on Discord? My handle is `thymikee#1279`, we should create a conversation for this preset :D

PR #290, opened 2019-07-11
fix: jest-preset-angular version in package.json

capitantrueno:


thymikee:
Example app is used for testing, this is intended. Feel free to change it locally if you need to :)

capitantrueno:
Thank you for your response. I thought it was a mistake.

Issue #286, opened 2019-06-18
Issue with Angular 8: `src/tsconfig.spec.json` => `tsconfig.spec.json`


yjaaidi:
Hi there,

Since Angular 8, `tsconfig.spec.json` isn’t in ˋsrc` anymore but in the parent folder.
Shall we change the default path in order to avoid overriding it in the configuration?

Cf. https://github.com/briebug/jest-schematic/issues/12

Thanks

wtho:
I personally agree, as we want to make the use of jest for Angular users as easy as possible.

PRs are welcome!

This will be a breaking change and should be documented in README.md and CHANGELOG.md.

yjaaidi:
Thanks @wtho for your quick reply.
First, I've got to make sure that migrating an Angular project from 7 to 8 moves the tsconfig.spec.json to the parent directory.
I'll send a PR as soon as I can. If anyone wants to work on this issue, please let me know.

sfabriece:
my migration to 8 didn't move it to the parent @yjaaidi

wtho:
Yeah, they only do this in newly generated Angular8 projects.

As we will support v7 and older as well as migrated projects, we will have to explain this in the migration guide.

I have not looked into the original change in the Angular repo, there might be a discussion that can help us to decide what to put as the default. I personally would interpret this change as that the recommended location now is in `/`, whereas before it was in `/src`. Anyway you can still configure it to be wherever you want, for Angular and jest.

mnicic:
As temporary solution, you can add this configuration into `jest.conf.js` file:
```javascript
  globals: {
    'ts-jest': {
      tsConfig: '<rootDir>/tsconfig.spec.json'
    },
  },
```

dgroh:
> As temporary solution, you can add this configuration into `jest.conf.js` file:
> 
> ```js
>   globals: {
>     'ts-jest': {
>       tsConfig: '<rootDir>/tsconfig.spec.json'
>     },
>   },
> ```

Now I get this error:

> TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics`

dgroh:
This now should work:

```js
module.exports = {
    preset: 'jest-preset-angular',
    setupFilesAfterEnv: ['<rootDir>/setupJest.ts'],
    globals: {
        'ts-jest': {
          tsConfig: '<rootDir>/tsconfig.spec.json',
          diagnostics: false
        },
      },
};
```

[...]

PR #299, opened 2019-08-07
Update README on how to override default values

basickarl:
Currently when trying to override the global ts-jest field with custom values,
the README.md points towards the local configuration that the module uses. This
version of the configuration will not work if the configuration is in any other
location other than in the preset module.

Related: https://github.com/thymikee/jest-preset-angular/issues/286#issuecomment-519050423

For a user to override this part of the object the user must specify the location
of the preset module for it to resolve correctly.

Added a clear example so users will know how to override preset configuration
values and mix them with normal Jest configuration values.


basickarl:
Just realized this is stated in the area: https://github.com/thymikee/jest-preset-angular#custom-tsconfig

PR #307, opened 2019-09-14
feat: restructure project with src/ and build/

wtho:
### Problem
More and more js files such as serializers were located in the root of the project separated from ts src files, while these ts files also got transpiled to the root folder. This creates a non-transparent file mix with compiled and source files in the same place.

### PR
To keep source files in a single place, most js files are now located in the src folder as well. A build folder will be created on `npm run build`, `yarn build` or `npm run prepare`, as configured in `tsconfig.json`.

The TypeScript compiler transpiles the ts files and conventiently copies the js files to the build folder. This means the js files are not changed by the tsc.

I kept the `index.js` and `jest-preset.js` in the root folder, as otherwise the preset would not be referenceable as just `jest-preset-angular` from the jest config.

All other js files, such as transformers and serializers have to be referenced as `jest-preset-angular/build/...` now. This makes this change a breaking change, as many users will have to adjust their configuration.

On the other hand this makes it much more easy for us maintainers and new contributers to work with the source files, as the project setup is more intuitive.

The naming of `build` was inspired by `facebook/jest`.


README.md
```diff
@@ -411,7 +417,7 @@ module.exports = function(api) {
 {
   "jest": {
     "transform": {
-      "^.+\\.(ts|html)$": "<rootDir>/node_modules/jest-preset-angular/preprocessor.js",
+      "^.+\\.(ts|html)$": "ts-jest",
```
wtho: Not sure about this, does this apply for a config with babel + ts-jest? Surely the referenced `preprocessor.js` would not work anymore.

ahnpnl: I remember when using with babel, people often specify pattern to use babel here to deal with js and pattern ts with ts-jest at this place, here is a [reference](https://stackoverflow.com/questions/52173815/ts-jest-does-not-recognize-es6-imports)

wtho: Ok cool, so if I understand it properly, forwarding only `*.ts` to `ts-jest` should do the trick here, as done in the example. Do you agree?

ahnpnl: ya I think that should work. In the past we tried to override ts-jest preprocessor so that was why we needed to point to our `preprocessor.js`. Since now we don't have that anymore so we can use `ts-jest` and `babel-jest` like the example.

ahnpnl: and I think we should think about a way to opt-in `ts-jest` for `*ts.` files instead of forcing users to use it. Now babel has plugin for const enum as well, so when using babel the users only lose the type-checking functionality of `tsc`. 

In my opinion, we should give a free choice for users related to `babel` or `ts-jest` they want to use for `*.ts` files

wtho: Ok, so an alternate setup with a babel would be great! Maybe running the example project twice?

That's a task for a different issue though.

README.md
```diff
@@ -305,14 +308,17 @@ Override `globals` object in Jest config:
...
+          "jest-preset-angular/build/StripStylesTransformer"
+        ],
       }
     }
   }
 }
 ```
 
-If you choose to overide `globals` in order to point at a specific tsconfig, you will need to add `"astTransformers": ["jest-preset-angular/InlineHtmlStripStylesTransformer"]` to the `globals.ts-jest` section too, otherwise you will get parse errors on any html templates.
+If you choose to overide `globals` in order to point at a specific tsconfig, you will need to add the `astTransformers` to the `globals.ts-jest` section too, otherwise you will get parse errors on any html templates.
```
thymikee: How about linking to the specific lines in the source file?

__tests__/CoreJS-es6-reflect.test.ts
```diff
@@ -24,21 +24,21 @@ describe('importing ES6 reflect', () => {
     jest.mock('core-js/es6/reflect', () => { throw coreJs2Error }, { virtual: true });
     jest.mock('core-js/es/reflect', () => { throw coreJs3Error }, { virtual: true });
 
-    expect(() => require('../setupJest')).toThrow('core-js es6-reflect not found!');
+    expect(() => require('../src/setupJest')).toThrow('core-js es6-reflect not found!');
```
thymikee: let's move this test into `src/` instead :)

__tests__/CoreJS-es7-reflect.test.ts
```diff
@@ -24,21 +24,21 @@ describe('importing ES7 reflect', () => {
     jest.mock('core-js/es7/reflect', () => { throw coreJs2Error }, { virtual: true });
     jest.mock('core-js/proposals/reflect-metadata', () => { throw coreJs3Error }, { virtual: true });
 
-    expect(() => require('../setupJest')).toThrow('core-js es7-reflect not found!');
+    expect(() => require('../src/setupJest')).toThrow('core-js es7-reflect not found!');
```
thymikee: same as above

__tests__/InlineFilesTransformer.test.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import * as tsc from 'typescript';
-import * as transformer from '../InlineFilesTransformer';
+import * as transformer from '../src/InlineFilesTransformer';
```
thymikee: same as above

__tests__/StripStylesTransformer.test.ts
```diff
@@ -5,7 +5,7 @@
  */
 
 import * as tsc from 'typescript';
-import * as transformer from '../StripStylesTransformer';
+import * as transformer from '../src/StripStylesTransformer';
```
thymikee: same as above

src/setupJest.ts
```diff
@@ -25,7 +25,7 @@ require('zone.js/dist/proxy.js');
 require('zone.js/dist/sync-test');
 require('zone.js/dist/async-test');
 require('zone.js/dist/fake-async-test');
-require('./zone-patch');
+require('../zone-patch');
```
thymikee: `zone-patch` should also be in `src`

thymikee:
👍

README.md
wtho: The *Custom tsconfig* with the full example part is just right above this line, I think that is easy enough to copy. If we link the full preset, they might copy wrong parts, like `"astTransformer": ["./build/Transformer"]`, so I actually prefer this style.

Plus the configuration is linked on the very top of the README.

wtho: image

__tests__/CoreJS-es6-reflect.test.ts
wtho: I just put it inside `src/`: `src/CoreJS-es6-reflect.test.ts`, better have a look at the folder structure in this branch.

Is this ok? The `__snapshot__` folder also lives inside `src/` now. Or do you think we should separate them somewhat from the source files?

IMO with the relative small amount of source files it is ok to keep them in one folder for now.

ahnpnl: Hmm why don't we leave test files in src/ __tests__  ? I think it's better and it's a pattern used by each jest package folder too

README.md
ahnpnl: 👍

__tests__/CoreJS-es6-reflect.test.ts
wtho: You are right! I argued with being inspired by the facebook/jest setup!

Would you put the `tsconfig.spec.json` for the preset tests inside there too?

ahnpnl: I think that makes sense 👍

README.md
wtho: Opened #308

.npmignore
```diff
@@ -1,4 +1,3 @@
 /example
-/__tests__
```
thymikee: Should be changed to just `__tests__` or better, let's use `files` entry in package.json to distribute only what we want

package.json
```diff
@@ -31,7 +31,7 @@
         "tsConfig": "<rootDir>/tsconfig.spec.json"
       }
     },
-    "rootDir": "__tests__"
+    "rootDir": "src/__tests__"
```
thymikee: can we remove custom rootDir at all? feel free to do it later

tsconfig.json
```diff
@@ -11,5 +11,5 @@
     "types": ["node"]
   },
   "include": ["src"],
-  "exclude": []
+  "exclude": ["**/__tests__"]
```
thymikee: do we need to exclude tests?

.npmignore
wtho: I removed this line, as it's inside the `src/` now anyway.

thymikee: Ah yea, didn't notice we ignore src

package.json
wtho: We need it for `ts-jest` to find `tsconfig.spec.json`. We could also configure the location of the tsconfig it in `globals.ts-jest` in `package.json` instead.

tsconfig.json
wtho: Yes, otherwise `tsc` will also compile `src/__tests__/**` to `build/`

Issue #288, opened 2019-07-03
Angular 8: Can't resolve all parameters for Component: (?).


jfcere:
I recently upgraded my application to Angular 8 and couldn't get the tests to run successfully with _Jest-Preset-Angular_ resulting in the error below telling me it cannot resolve the constructor dependencies although it works fine using _Karma_ and was working with Angular 7.

```
Can't resolve all parameters for AppComponent: (?).
```

I've created a newly generated Angular 8 application to see if it has something to do with my migration and I ended up having the same issue.

Here is the reproduction repository: https://github.com/jfcere/jest-preset-angular-issue

wtho:
### TL;DR
```diff
tsconfig.spec.json

   "compilerOptions": {
+    "emitDecoratorMetadata": true,
     "outDir": "./out-tsc/spec",

```

### Issue
The Angular CLI takes over some of the TS -> JS transpilation using transformers, especially when it's about the decorators and reflection. We were not aware of this change, but it appearently came in in Angular v8.1.

Here the full PR: https://github.com/angular/angular-cli/pull/14473

We can try to reuse their transformer (https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/transformers/ctor-parameters.ts), as `@angular/cli` is anyway a dependency in every Angular project.

I will try to experiment with the transformer, we can track the progress in this issue till then.

### Workaround
To make `ts-jest` not lose the parameter information upon compilation without the AST transformation, you can set `"emitDecoratorMetadata": true` in your `tsconfig.spec.json` for now. The PR mentions issues with AOT compilation, which is usually not used in testing, so this setting should not create any problems. If you use karma as well and this leads to issues, try to use a separate `tsconfig.spec.json`.

jfcere:
@wtho thanks for the quick answer, really appreciated!

I've downgraded the reproduction repository to use Angular 8.0.3 and the error is still happening.

But the workaround for `tsconfig.spec.ts` you've provided works! 😄

wtho:
Ok interesting, I created a project with 8.0.0 once and there `emitDecoratorMetadata` was still set, I thought it would not have come in in a patch version update, but for Angular it is just an internal detail.

It's difficult to spot, as I could not find the PR referenced in the [angular-cli releases list](https://github.com/angular/angular-cli/releases) anywhere.

jfcere:
>  I created a project with 8.0.0 once and there emitDecoratorMetadata was still set

[...]

Issue #266, opened 2019-05-15
Angular no longer directly depends on core-js - should not be a peerDependency


okunokentaro:
Hi there.

https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/workspace/files/package.json.template#L22-L24

Currently, Angular CLI does not require core-js and is not installed. On the other hand, `jest-preset-angular` has `peerDependencies` dependencies in `core-js`.

It seems to me that we will be forced to do unnecessary installations and should be removed dependencies on core-js. Thanks.

wtho:
According to the commit message https://github.com/angular/angular-cli/commit/a0a583adaaf8e739c375b914f249801660f33334, the "Angular Build System takes care of the dependencies".

As jest does not use the Angular Build System and requires some library to reflect the metadata in the decorators, my first suggestion would be to make it a dependency.

Do you know how Angular handles the reflection lib? Does it install it on-demand at the first build? Does Angular offer to use other libraries than `core-js`?

wtho:
I just installed the latest anguar cli and created a brand new project. After that, the package.json still included `core-js`.

As we will need some metadata reflection, I think this is no option to remove the dependency.

Did you manage to create a project with the cli that did not install `core-js`? If so, can you give instructions to replicate this behavior?
Also some more information on how Angular installs the metadata polyfill on-demand or if it does the reflection in some other way would be required to move this issue forward.

lacolaco:
@wtho Hi. Angular CLI inserts reflect-metadata polyfill in the phase of build without `--aot` flag. It means the CLI requires `core-js` in `node_modules` for that behavior. 

`core-js` is now a part of `dependencies` of `@angular-devkit/build-angular`, not _peer_.  
https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/package.json#L22
So, Angular CLI no longer add `core-js` in the application's package.json.  

I think jest-preset-angular cannot run without reflection-metadata. And it can depend on `core-js` as not a peer, like Angular CLI. I think it is better than asking for additional installation to users.

The key is there is no longer guarantee the Angular applications have `core-js`.

lacolaco:
That feature, adding JIT polyfill, is landed from v8. Current `latest` CLI doesn't support it. Try to use `@angular/cli@next` to confirm that!

wtho:
Yeah, I can confirm it is located inside `@angular-devkit/build-angular`.

[...]

PR #311, opened 2019-09-18
feat: Bump example project to Angular 8

wtho:
I did most adjustments to be in sync with the default Angular CLI Boilerplate. I know many Angular projects will still differ from this, but I feel like it is the default starting point in the community. Open to discussion regarding this.

Changes in example project:
* Removed the whole e2e section in `angular.json`, as it looks quite different in current generated projects, but updating also the e2e folder seemed out of scope.
  **Question:** Should we remove the e2e source folder as well? Or keep everything also in `angular.json` in sync to the boilerplate of angular 8?
* I am not sure about the dependencies. Angular projects now lock the versions inside `package.json` using `~x.x.x`.
  **Question:** Should we keep all versions locked in the example project as well?
* Removed `core-js` from peerDependencies, it is not a dependency of Angular projects anymore, and made it a dependency of `jest-preset-angular`, as we need it for testing. See #266 for further discussion.
* Replaced `HttpModule` with `HttpClientModule` (`HttpModule` was dropped completely)
* `tsconfig.json` `baseUrl` is now the root folder, as it is in the Angular CLI boilerplate project
* `tsconfig.spec.json` needs `emitDecoratorMetadata: true` (or access to program/type checker through ts-jest), see #288 
* Adjustment to make `ng build --aot` not fail: ```[\`styles\`]: ...``` had to become `styles: ...` again and declared all components to `app.module`
* Updated `polyfills.ts` to current one
* Updated all ts config files
* Updated ts lint config file

Note that most tested files and tests remained untouched. Mostly the Angular Project files were changed.
Is this breaking due to moving `core-js` to dependencies?

Renaming the example project to e. g. `test-app` would be another task I would like to start with after this PR.

I would like to keep #288 open, as the cleaner approach might be a transformer (one day).

Solves #266.


wtho:
Just realized a changelog entry is still missing. Will add it soon!

ahnpnl:
- I agree e2e should be removed. It is irrelevant for the example app.
- I think the versions of dependencies for test app can just be kept in package.json as they are now.
- I think now we include core-js in dependencies is a breaking in the meaning of preset scope. In the meaning of usage scope, it's not really a breaking since it doesn't break anything on users' side unless users uninstalled core-js (I don't think there are any polyfill libraries better than it)

wtho:
There are polyfill libraries that are way more lightweight than `core-js` for the decorator metadata, but as it was a dependency of Angular for such a long time, we should still stick to it. Especially when the preset will be used with an older Angular version the users would have `core-js` installed already.

wtho:
Last Force Push
* Added Changelog information
* Removed the e2e folder and e2e configuration file
* Updated basic information of the example app `README.md` to inform users what the example app is

CHANGELOG.md
```diff
@@ -1,15 +1,19 @@
...
+* (**BREAKING**): Remove `core-js` peer dependency and instead add it as direct dependency ([#311](https://github.com/thymikee/jest-preset-angular/pull/309)). 
+
+#### Chore && Maintenance
+* Update example app to match Angular 8 Boilerplate ([#311](https://github.com/thymikee/jest-preset-angular/pull/309)).
 
 #### Migration Guide
 * If the `astTransformers` are referenced in a custom `jest` config, `[ 'jest-preset-angular/build/InlineFilesTransformer', 'jest-preset-angular/build/StripStylesTransformer']` have to be set instead.
 * Serializers, transformers and `setupJest` have to be referenced from the `jest-preset-angular/build/`-folder in a custom config. Existing references have to be aligned.
 * If your `tsconfig.spec.json` is located in `src`, move it to your root folder and adjust the referenced files and paths inside, or align your jest configuration as discussed in the [README.md](https://github.com/thymikee/jest-preset-angular/blob/master/README.md#custom-tsconfig).
+* In an Angular 8 project or higher, `core-js` should no longer a dependency in the `package.json` of the Angular project, except if it is imported manually in the code.
```
thymikee:
```suggestion
* In an Angular 8 project or higher, `core-js` should no longer be a dependency declared in the `package.json`, unless it's imported manually in the code.
```

example/tslint.json
```diff
@@ -1,36 +1,49 @@
 {
-  "rulesDirectory": [
-    "node_modules/codelyzer"
-  ],
+  "extends": "tslint:recommended",
```
thymikee: lol, Angular still uses tslint? It's long been deprecated, we should use eslint instead, but if the default template comes with it, I'm ok

wtho:
Last Force Push:
* removed tslint dependency and configuration in the example app (was ignored and never run)
* rephrased changelog as recommended

PR #313, opened 2019-09-25
docs: add a warning message for exposed config

### LINKED ISSUE #312: Docs: add a temporary warning message in README about exposed config
The current exposed config in the README file shows the latest configuration that is not published to NPM yet.

If the new version isn't going to be release soon it could help to add a warning message to indicate that the configuration for the latest NPM version varies and should be retrieved from a different place.

ihadeed:
Closes #312

The message can be hidden on npm by using templates, or a quick `prepublish` script that runs an `sed` command (or equivalent) so it shows up on Github only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ By Angular CLI defaults you'll have a `src/test.ts` file which will be picked up
 
 
 ## Exposed [configuration](https://github.com/thymikee/jest-preset-angular/blob/master/jest-preset.js)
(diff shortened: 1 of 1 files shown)


wtho:
Thanks, good idea!

I think we will just remove it when we publish the next version.
Usually there are no big changes, it's just that this time several breaking changes are landing at the same time, and it takes a while.

ihadeed:
>  it's just that this time several breaking changes are landing at the same time, and it takes a while.
Totally understandable :smile: 

Thank you for your great work - being able to just plug-and-play jest into my Angular apps is a big time saver.

PR #317, opened 2019-10-02
feat: alternative setup with babel/typescript

### LINKED ISSUE #308: Alternate setup with `@babel/preset-typescript` without `ts-jest`
Babel put quite some resources into transpiling TypeScript to JS as well, and with the first-class support of babel in jest we would like to give the user the choice of using `ts-jest` or babel for transpiling their angular projects for jest.

Setting a babel option up and providing this to users will make this preset less opiniated and more objective, plus will also give us a clear picture which mode is faster and what problems arise with what transpilers respectively.

-----

There are known caveats though, as babel does not type check the tests itself (similar to `ts-jest`'s [`isolatedModules: true`](https://kulshekhar.github.io/ts-jest/user/config/isolatedModules)).

wtho:
⚠️ So far this is only a POC to show babel + jest does work without `ts-jest` for Angular projects, this serves as a basis of discussion.

This would add an alternative compiling method alongside `ts-jest`.

Done so far
* Babel config is in plain js inside new `babel` folder
* second preset is also located in there
* added  second preset inside `babel` folder, which can be used via `preset: 'jest-preset-angular/babel'` in jest config
* added basic preconfigured `babel.config.js`
* added `load-html-transformer.js` to deal with `html` files
* added `inline-template.plugin.js` to do what `InlineFilesTransformer.ts` does
* removed `ts-jest` from dependencies, which means it would have to be be installed when using this preset with `ts-jest` from now on

The following libraries have to be installed alongside the project, when setting up jest-preset-angular + babel
* `@babel/core`
* `@babel/preset-env`
* `@babel/preset-typescript`
* `@babel/plugin-proposal-decorators`
* `@babel/plugin-proposal-class-properties`
* `babel-plugin-transform-typescript-metadata`

What is still missing for this to land in the preset?
* separated folder organization for config files of `ts-jest` vs `babel`
* implementing babel js files in ts whenever feasible
* Unit Tests for the plugin/html-loader
* renaming of babel plugin and html loader files from `kebab-case` to `PascalCase`
* Tests for example app that does run with both, `ts-jest` and `babel-typescript`
* Extensive documentation in README.md

-------

 This would be especially interesting regarding some feedback from the community about speed difference in bigger projects. Babel does no TypeChecking in the `.test.ts`/`.spec.ts` files itself, but going through the issues of `ts-jest` (e. g. https://github.com/kulshekhar/ts-jest/issues/1115, surely ahnpnl knows more) some projects are falling back to `isolatedModules: true` already, which is also skipping the type checking. Furthermore type checks can still be achieved separately using `tsc --noEmit`.

Personally I think we need some more restructuring in the preset to keep `ts-jest` and `babel` separated as two exclusive usage options, so this might take some time. I think this can land in v9 later this year.

Closes #308


ahnpnl:
I think we need also plugin for babel plugin const enum for this PR

wtho:
Yeah! Let's add a test with enums in the example app!

wtho:
**In last force push:**
* Added a test with `enum` and `const enum` and added the babel plugin for const enums to ensure the preset can deal with this TS language feature when using babel.

wtho:
In the newest push I added some opiniated changes, so feel free to correct it wherever it looks seldom.

**Proposal: Folder Structure with babel & ts-jest**
```
.
├── README.md
├── package.json
├── [other-base-level-files]
└── src
    ├── AngularNoNgAttributesSnapshotSerializer.js
    ├── AngularSnapshotSerializer.js
    ├── HTMLCommentSerializer.js
    ├── index.ts
    ├── reflectMetadata.ts
    ├── setupJest.ts
    ├── babel
    │   ├── babel.config.ts
    │   ├── jest-preset.ts
    │   ├── load-html-transformer.ts
    │   └── plugin-inline-angular-template.ts
    ├── ts-jest
    │   ├── InlineFilesTransformer.ts
    │   ├── jest-preset.ts
    │   ├── StripStylesTransformer.ts
    │   └── TransformUtils.ts
    ├── __tests__ [folder with tests]
    └── zone-patch [folder with sources]
```
`ts-jest` and `babel` would be built to `jest-preset-angular/build/ts-jest` and `jest-preset-angular/build/babel` respectively.

This would lead for users to adjust their `jest.config.js`:
```js
module.exports = {
  preset: "jest-preset-angular/build/ts-jest",
  // or
  preset: "jest-preset-angular/build/babel"
}
```
We could add additional folders with a proxy `jest-preset.js` at `jest-preset-angular/babel` and `jest-preset-angular/ts-jest`, but that would add additional confusion to our project itself. Anyway there should be a `jest-preset.js` in the root that points at `build/ts-jest/jest-preset.js` for backwards compatibility.

-----

**Proposal: Run project example app with ts-jest and babel**
To run the project with both the settings, I suggest we set the `preset` value in `example/jest.config.js` programmatically, e. g. depending on a env variable:
`export JPA_TEST_MODE=ts-jest yarn test`
```js
// jest.config.js
const testMode = process.env.JPA_TEST_MODE || 'ts-jest';

module.exports = {
  preset: `jest-preset-angular/build/${testMode}`,
  ...
```

This way we do not have to duplicate any code base and make sure the two transpilation modes are actually easily exchangable. Both transpilation plugins from babel and ts-jest have to be installed initially, which would also speed up the CI test time.

There is one issue (see CI tests): `tsc` transpiles undefined class properties differently than `babel/typescript` does (babel adds the property):
```diff
  <app-calc
    hasAClass="false"
    image={[Function String]}
+   observable$="undefined"
    prop1={[Function Number]}
  >
```

We have to plugin optional tests somehow. Any experience with this?

-----

This PR would be a big difference for this preset, so it is quite important to me what you two think, @ahnpnl and @thymikee. Note that the speedup of the example app testrun is similar to the one of `ts-jest` with `isolatedModules: true` (~4s). Would be interested if there is any difference in bigger projects though!

ahnpnl:
I like the idea of having 2 internal presets for babel and ts-jest and let users choose what they want to use. 
Regarding to the folder structure, all LGTM but imo I like putting __tests__ outside src folder. It will be easier for us to build and release (less configuration in excluding something).

I haven't tried to use babel yet so I wasn't aware of the different in compiling. Looks interesting. Is it intended behavior of babel team ?

wtho:
Now it is definitely ready for review, although I do not feel good about the test mode detection. Maybe you have a better idea?

Also with [this test](https://github.com/wtho/jest-preset-angular/blob/feat/babel-jest-angular/example/src/app/calc/calc.component.spec.ts#L25-L55) that behaves differently (`observable$="undefined"`) I am not sure if my inline snapshot is a good solution.

Good thing is I used the very same test cases for the babel ast transformer as for the ts-jest ast transformer. I duplicated them for now, but can extract them to a single source.

ahnpnl:
> Now it is definitely ready for review, although I do not feel good about the test mode detection. Maybe you have a better idea?

I'd do similar things as well. 

> Also with [this test](https://github.com/wtho/jest-preset-angular/blob/feat/babel-jest-angular/example/src/app/calc/calc.component.spec.ts#L25-L55) that behaves differently (`observable$="undefined"`) I am not sure if my inline snapshot is a good solution.

Maybe use `fixture.nativeElement` instead of `fixture`. I think testing HTML template shouldn't use that `fixture` but rather than HTML elements rendering. And for the different output related to `ts-jest` vs `babel` compiling, I think there should be a note highlighted in the documentation about that.

package.json
```diff
@@ -2,20 +2,26 @@
...
-  "main": "index.js",
+  "main": "build/index.js",
   "repository": "git@github.com:thymikee/jest-preset-angular.git",
   "homepage": "https://github.com/thymikee/jest-preset-angular",
   "author": "Michał Pierzchała <thymikee@gmail.com>",
   "license": "MIT",
   "dependencies": {
     "jest-environment-jsdom-fifteen": "^1.0.0",
-    "pretty-format": "^24.0.0",
-    "ts-jest": "^24.0.0"
```
ahnpnl: since `ts-jest` is no longer installed automatically by this preset anymore, we should add this to the breaking part of `CHANGELOG`

wtho: Done!

wtho:
> Maybe use `fixture.nativeElement` instead of `fixture`. I think testing HTML template shouldn't use that `fixture` but rather than HTML elements rendering.

Snapshotting `fixture.nativeElement` does solve the issue, as it does not render the properties, but I guess that was the point of the original test. The test then will diff this:
```diff
- <app-calc
-   hasAClass="false"
-   image={[Function String]}
-   prop1={[Function Number]}
+ <div
+   id="root0"
  >
    <p
      class="a-default-class"
    >
       calc works! 1337 another text node test-image-stub 
    </p>
- </app-calc>
+ </div>
```
I am not so sure if that's a good snapshot anymore.

ahnpnl:
The snapshot thing indeed depends on the purpose of users when testing a component. If users care about the property rendered by ts file, fixture is the right one. Otherwise if users only care about html change, fixture.nativeElement is the choice.

I think we don't need to care about this too much. Perhaps we can add a note in README to warn about the difference of snapshot relative to usage of tsc vs babel. I think that should be fine.

At work I usually use fixture.nativeElement because I only want to check about html change.

wtho:
👍  Already added a small note in the [README.md](https://github.com/thymikee/jest-preset-angular/pull/317/files#diff-04c6e90faac2675aa89e2176d2eec7d8R43). I know, it's not very visible, but someone who does read through the README will encounter it.

BensonBen:
@thymikee normally, I don't comment on pull requests. But I'm pretty interested in using this if it leads to test suite speed increases, so your feedback is valued.

For now, I'm actually going to be converting my teams jest tests to use babel. As my initial testing lead to some serious speed improvements with ~400 individual tests. 

@wtho I can report rough performance improvements if you'd like. Also, I'm using the `isolatedModules: true` option in our tsconfig.

thymikee:
Making it easier to use Babel sounds like a good idea, I just didn't find the time to have a look yet :D

src/babel/load-html-transformer.ts
```diff
@@ -0,0 +1,10 @@
+import { Transformer } from "@jest/transform";
```
BensonBen: Consider making this single quotes for consistency in what I've seen so far in this PR

wtho:
@BensonBen Yeah, a performance evaluation would be super helpful! I do not have a big codebase at hand currenlty. Maybe you can install jpa using `npm install -D git://github.com/wtho/jest-preset-angular.git#feat/babel-jest-angular` and run them with the new configuration.

Also I'm interested in which adjustments were required to use babel. Making switching as easy as possible would be desired from our side.

src/babel/load-html-transformer.ts
wtho: Done

wtho:
True, we can keep the structure and just point it at ts-jest as default. I just liked the equality between the two options and disregarded backwards-compatibility.

I would also suggest to create a new issue to gather babel feedback and link the issue next to the babel setup notes.


**Changes in last commit**
* add default `jest-preset.js` to provide full backwards-compatibility
* add `ts-jest` to dependencies
* Set package version to 8.1.0 (ready for squash, tag & push)
* adjust CHANGELOG.md
* rephrase README.md, add notes for preset comparison feedback (can be removed later again)

ViceIce:
I would like to test, but i get errors with babel

*ts-jest*
image

*babel*
I get errors like:  `Can't resolve all parameters for ...: (?).`

image

babel is a lot slower too, using following additional settings for `ts-jest`
```js
diagnostics: false,
isolatedModules: true,
```

wtho:
Interesting!

Can you create a gist just containing the classes that have injection issues (and the classes/providers that could not be injected)?
Decorators, classname and constructor signatures are sufficient. Methods, class fields and templates are not involved, so better remove them.

This information would help a ton!

ViceIce:
looks like it is alwys the same class `MyHandler` which is used to provide a service.
I hope this is enough to reproduce.

Technically I'm using a `nrwl` workspace with multiple library packages.

```ts
import { NgModule, Injectable, Optional } from '@angular/core';
// import { NGXLogger } from 'ngx-logger';
// import { MessageService } from 'primeng/api';
import {
    MissingTranslationHandler,
    MissingTranslationHandlerParams,
    TranslateCompiler,
    TranslateLoader,
    TranslateModule,
    TranslateService,
} from '@ngx-translate/core';

@Injectable({ providedIn: 'root' })
export class Logger {
  constructor(/* private readonly _log: NGXLogger */) {}
}

@Injectable()
export class MessageService {
}

class MyHandler {
  constructor(@Optional() private messageService: MessageService, private readonly _log: Logger) {}
  handle(params: MissingTranslationHandlerParams) { return params.key; }
}

@NgModule({
  imports: [ TranslateModule ],
})
export class SharedModule {
constructor(ts: TranslateService, private readonly _log: Logger) {}
}

describe('SharedModule', () => {

	beforeEach(async(() => {
		TestBed.configureTestingModule({
		  imports: [
			TranslateModule.forRoot({
			  missingTranslationHandler: { provide: MissingTranslationHandler, useClass: MyHandler },
			  useDefaultLang: false,
			}),
		  ],
		  providers: [{ provide: MessageService, useValue: { add: jest.fn() } }],
		}).compileComponents();
	});
	
	it('should create', () => {
        expect(SharedModule).toBeDefined();
    });
});
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Changelog (master)
 
+### v8.1.0
+
+#### Features
+* Add preset `babel` as transpilation alternative to `ts-jest` ([#317](https://github.com/thymikee/jest-preset-angular/pull/317)).
+
+#### Migration Guide
+* The preset reference in the jest config has to be updated from `"jest-preset-angular"` to `"jest-preset-angular/build/ts-jest"`. To use babel, replace `ts-jest` with `babel`. Furthermore, when using `ts-jest`, it has to be added as `devDependency` to the project, or `babel` packages alternatively (see README.md).
```
thymikee: This is not relevant now

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Changelog (master)
 
+### v8.1.0
+
+#### Features
+* Add preset `babel` as transpilation alternative to `ts-jest` ([#317](https://github.com/thymikee/jest-preset-angular/pull/317)).
```
thymikee:
```suggestion
* Expose a new preset `jest-preset-angular/build/babel` for transforming files with Babel instead of `ts-jest` ([#317](https://github.com/thymikee/jest-preset-angular/pull/317)).
```
Side note: would be nice if we could get rid of `build` part, just `jest-preset-angular/babel` – easy enough with re-exporting from top-level.

README.md
```diff
@@ -19,6 +22,19 @@ npm install -D jest jest-preset-angular @types/jest
 
 This will install `jest`, `@types/jest`, `ts-jest` as dependencies needed to run with Angular projects.
 
+Additionally you need `babel` packages if you want to use `babel`.
```
thymikee: Please remove, no need to mention it here
```suggestion
```

README.md
```diff
@@ -19,6 +22,19 @@ npm install -D jest jest-preset-angular @types/jest
...
+
+### babel
```
thymikee: How about:
```suggestion
### jest-preset-angular/babel
```

README.md
```diff
@@ -19,6 +22,19 @@ npm install -D jest jest-preset-angular @types/jest
...
+`babel` uses the babel compiler to generate JavaScript, without type-checking your test files before doing so. This might result in a different test run performance. You can still run type-checking using `tsc --noEmit`. For additional TypeScript-supported language features you might have to install more babel packages. Note that `babel` also can differ slightly from `tsc`, e. g. in compiling a class to a function.
```
thymikee:
```suggestion
`babel` preset uses the Babel compiler to generate JavaScript without type-checking your test files before doing so. This might result in a different test run performance. You can still run type-checking using `tsc --noEmit`. For additional TypeScript-supported language features you might have to install more babel packages. Note that `babel` also can differ slightly from `tsc`, e.g. in compiling a class to a function.
```

CHANGELOG.md
wtho: I agree. Maybe we can just publish the `build` folder with the `package.json`? Have to check out how other projects solve this.

wtho:
@ViceIce I tried out the tests and as soon as I add `@Injectable` to `MyHandler` it works in babel, but not in ts-jest, so this seems contradictary to what you said.

Maybe we can figure it out in this minimal repro: [wtho/jpa-babel-bug](https://github.com/wtho/jpa-babel-bug/issues/1)

ViceIce:
@wtho Next bug wtho/jpa-babel-bug#2

ViceIce:
found workarounds for the bugs, here are the new timings:
image

second run 
image

wtho:
While we have to wait for https://github.com/leonardfactory/babel-plugin-transform-typescript-metadata/pull/24, you guys can help us by doing further tests on bigger codebases.

Please try out the babel preset using a branch where I included the babel-plugin PR and the built js files (which would not be included normally checking out a git branch as npm dep).
```
npm i -D jest-preset-angular@git://github.com/wtho/jest-preset-angular#test/tryout-babel-jest-angular
```
(make sure you followed the other instructions in the [jest-preset-angular/babel README](https://github.com/wtho/jest-preset-angular/tree/feat/babel-jest-angular#usage))

Please report any problems. Feel free to open issues on https://github.com/wtho/jpa-babel-bug

Thanks!

ViceIce:
News: Angular 9 is out and they require `@Injectable` more aggressive.

wtho:
Finally!
I think that's no problem for the babel scenario (or the preset in general).

ViceIce:
Yes, but fixes one of my bugs by design. 😊

ahnpnl:
FYI: https://github.com/kulshekhar/ts-jest/pull/1549 will be in alpha version of ts-jest (possibly today). You can test the alpha version and give some feedbacks for https://github.com/kulshekhar/ts-jest/issues/1115

ahnpnl:
I guess we can close this issue since we have to use `Program`

wtho:
I think we can keep open just in case this gets interesting again. I'll make it a draft though.

PR #318, opened 2019-10-02
chore: jsdom15 and enable AngularNoNgAttributesSnapshotSerializer by default

wtho:
* update default jsdom enviroment to v15
* enable AngularNoNgAttributesSnapshotSerializer by default
* use better reference of astTransformers (always in preset instead of
relative) to avoid user confusion in README


ahnpnl:
I think we should add a comment somewhere to remind ourselves to remove jest env jsdom 15 once jest 25 is out.

wtho:
Will v15 be default from jest v25 on?

Can you add a reference to that? Couldn't find anything in the jest repo. Thanks!

ahnpnl:
[Here ](https://github.com/facebook/jest/pull/8851)you go

wtho:
Adding a simple issue is a good way keep it in mind #325

wtho:
force-push: rebase master branch to fix conflicts

PR #327, opened 2019-10-21
ci(greenkeeper): greenkeeper update lockfile

wtho:
To avoid a new issue created by greenkeeper for each failing dependency update (like e. g. #250), the `greenkeeper-update-lockfile` program is used to handle the udpate in CI environments.

The [greenkeeper lockfile](https://github.com/greenkeeperio/greenkeeper-lockfile#circleci-workflows) documentation was followed to fix this. No changes on the codebase itself are necessary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,8 @@ jobs:
     executor: node-10
     steps:
       - checkout
+      - run: 'sudo yarn global add greenkeeper-lockfile@2'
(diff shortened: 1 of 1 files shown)


wtho:
I triggered it manually on my branch before creating the PR, maybe that's why circleci didn't run it again. Interesting that it does not link it though.

Anyway it passed on that branch.
I'm not sure if it actually fixes the issue with greenkeeper until greenkeeper has a new bump, but I guess we just have to try out.

PR #330, opened 2019-10-22
ci(release): add basic semantic-release setup

wtho:
This PR will enable automatic releases for every new versions of our package on npm. As you will see, this enforces a quite opiniated commit-style on us and our contributers, but would take work away from our side and bring CI to the next level:
* triggers release only from inside CircleCi
* triggers release only from master branch
* triggers release only if all tests pass

A release will
* figure out new version number
* update `package.json` and `CHANGELOG.md`
* push changed files to master
* publish new version on npm
* add release to GitHub with release notes

The release heavily depends on commit style conventions:
* Angular-style commit messages:  `<type>(<scope>): <subject> (#pr-nr)`, see the [Angular guidelines](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines) for more. Note that type "chore" was replaced "build", "ci" and "perf" and `<scope>` ~is not optional~ actually is optional.
* depending on commit type on the master branch, new releases will be triggered:
  * **Commit Bodies**, that start with (or contain) `BREAKING CHANGE:` will trigger a breaking release (major version upgrade)
  * `feat` will trigger a feature release (minor version upgrade)
  * `fix` and `perf` will trigger a patch release
  * other commits (like e. g. `ci(release): add basic semantic-release setup`) will not trigger any release
* If we want to slow down major version releases, we can avoid merging those on `master` and putting them on a `next` branch (as @thymikee pointed out)
* to reference PRs, they have to be included in the git message header
* a breaking change commit has to include a general git message, but also a breaking change note, which **only describes why it is breaking**.

See a full, optimal commit message:
```
feat(babel): add styleUrl transformer (#515)

BREAKING CHANGE: styleUrls were not handled before by the
babel astTransformer, but will now be output completely different.

Closes #510
```

This is a different convention as we used until now, where the `BREAKING CHANGE` label was just added to commit messages. See [this example](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-cli/CHANGELOG.md#200-2018-05-29) to understand how a single commit creates notes for the commit, but also for the breaking changes paragraph.

TODOs
- [X] Added `commitlint` and `husky` to make contributers follow commit message conventions
- [X] Added semantic-release/git to update changed files in repo
- [x] Create `contributing.md` with git message guidelines
- [x] Reformat existing changelog to conform with newly generated changelog (while changing as little as possible)
- [ ] Check if forked repos' master branch PR also triggers release
- [ ] Check if `semantic-release bot` can push on master (is it protected? If no, can anyone push to it currently?)
- [ ] Setup credentials for github and circleci token via semantic-release-cli (see [docs](https://semantic-release.gitbook.io/semantic-release/usage/getting-started), would you do it, @thymikee? I think this can wait though)

Every single bit of this PR is open for discussion.

-----

Edit 1: elaborate release, optimal commit and update TODOs


ahnpnl:
I also like the commit convention from the Angular team. At work I use commitlint in combination with husky to achieve the commit convention and also follow 90% of Angular team convention. 

I saw that in ts-jest CHANGELOG is generated for each PR which is merged into master but idk how exactly it is done.

In general I like the idea of putting into next branch/tag so we can keep master stable and release ready.

PR #331, opened 2019-10-26
License jest-zone-patch under MIT

wtho:
**Background**: `jest-zone-patch` was imported into this repository, as it seemed to only be used as transitive dependency of `jest-preset-angular` (https://github.com/thymikee/jest-zone-patch/pull/9#issuecomment-445772763)

-----

As discussed in #214, we would like all contributers of [`jest-zone-patch`](https://github.com/thymikee/jest-zone-patch/graphs/contributors) to agree for the `jest-zone-patch`-code to adopt the MIT license instead of the previous BSD-3 license.

Removing the second license from this repository will reduce confusion and complexity, I am also not sure if the second license is picked up by license collectors when bundling, or by dependency-license-checkers.

I am not sure if this procedure is required, but it is still nice to inform and ask the contributers.

- [x] @thymikee
- [x] @JLHwung 
- [x] @smithad15 (I assumed, as he liked the PR description)
- [x] @4kochi

A further step could be to archive [`jest-zone-patch`](https://github.com/thymikee/jest-zone-patch).


4kochi:
Nice move. I agree

JLHwung:
I am good to see it changes.

PR #334, opened 2019-11-28
Improve coverage defaults of preset

marklagendijk:
By default only files that are actually used in tests are included in the coverage report. This change ensures that files have no tests (0% coverage) are included in the report.

### DIFF
--- a/example/jest.config.js
+++ b/example/jest.config.js
@@ -11,5 +11,10 @@ module.exports = {
   },
   setupFilesAfterEnv: [
     "<rootDir>/src/setupJest.ts"
+  ],
+  collectCoverageFrom: [
+    "<rootDir>/src/app/**/*.ts",
+    "!<rootDir>/src/app/**/*.module.ts",
+    "!<rootDir>/src/app/**/test/*"
   ]
 }
--- a/example/jest.config.js
+++ b/example/jest.config.js
@@ -14,7 +14,6 @@ module.exports = {
   ],
   collectCoverageFrom: [
     "<rootDir>/src/app/**/*.ts",
-    "!<rootDir>/src/app/**/*.module.ts",
-    "!<rootDir>/src/app/**/test/*"
+    "!<rootDir>/src/app/**/*.module.ts"
   ]
 }


example/jest.config.js
```diff
@@ -11,5 +11,10 @@ module.exports = {
   },
   setupFilesAfterEnv: [
     "<rootDir>/src/setupJest.ts"
+  ],
+  collectCoverageFrom: [
+    "<rootDir>/src/app/**/*.ts",
+    "!<rootDir>/src/app/**/*.module.ts",
+    "!<rootDir>/src/app/**/test/*"
```
ahnpnl: where does this `test` folder come from ?

marklagendijk: I thought it was an Angular Style guide convention to have `test` folders which would contain test utilities. Turns out this is not the case.
Removed the exclusion

marklagendijk:
My idea was to adjust it here so people following the example app would see it, and use it.
But maybe it should be added to the preset itself (`jest-preset.js`?) instead? Or both?

wtho:
### Some Background:

The "example app" is a test app, where we hook up the preset in an angular app to test if it works as expected. It is just a part of our testing, an essential one. One could check how we set it up to get started with jest + angular configuration, but it is definitely not meant to be a boilerplate (as mentioned in the example app README).

So your PR improves the testing in our CI of this library. We have two test runs, executed by CircleCI:
* one for the preset unit tests and
* one for the test app with the preset

This PR currently adds this option to the one in the test app, now I also want to have it added to the preset unit tests, which are configured in `package.json` (see `jest`-section) to maintain consistency between the two test runs.

If you only want to add it so that other users can check it out, a better place might be the root `README.md`. Or you can submit another PR where the `example` app is renamed to `test-app` and after that a true `example` project is added as reference to this preset. But I think we are not so much interested in maintaining more code, but more focusing on issues with the preset itself.

marklagendijk:
@wtho thanks for the explanation.
So there are 3 jest configurations in the repository:
1. The preset: `jest-preset.js`
2. The unit tests of the library: `package.json`
3. The test app: `example/jest.config.js`

The configuration I added takes care that if you have a file with 0% coverage, the file _does_ show up in the coverage report. Without the configuration it would not show up, because the file wouldn't be used during the test run because it isn't imported in any `.spec.ts` file.

Based on this I would propose the following:
1. Add this configuration in the preset. Having it there would improve the base configuration of anyone using the preset.
2. Don't add it to the unit test config. It doesn't help much. The preset only contains a few files.
3. Don't add it to the test app. It will be in the preset already.

If you agree I will make these changes.

wtho:
Ok so this turns out to be a different feature request now.

To add this to the preset we have to:
* Reflect if this should be in this preset - factor in e. g. why there is no default setting in jest itself
* Evaluate what a good default is from an Angular+Jest-user perspective - how would they usually organize their tests?
* Decide if this change is breaking - what if someone made the coverage an essential part of their workflow/pipeline (e. g. min cov thresholds)
* add documentation

Maybe you can also adjust the issue title.

marklagendijk:
>  Reflect if this should be in this preset - factor in e. g. why there is no default setting in jest itself

I would say that the difference is that the Angular preset can assume more about the project. I would say this is in line with the `moduleNameMapper` setting of this preset.

> Evaluate what a good default is from an Angular+Jest-user perspective - how would they usually organize their tests?

There are a number of standard rules for this in Angular projects. Both documented in the Angular Style Guide, and applied by the Angular CLI:

- Applications files live under `/src/app`
- Test files next to the file they test, and are named [FILENAME.spec.ts](https://angular.io/guide/styleguide#unit-test-file-names)

Based on this I would say that the values I am currently proposing are good:

``` js
  collectCoverageFrom: [
    "<rootDir>/src/app/**/*.ts",
    "!<rootDir>/src/app/**/*.module.ts"
  ]
```

- All Typescript files under app are included.
- Except module / routing module files (these are generally not unit tested).
- So if you forget to create a `.spec.ts` file for your application file, it will be shown in the coverage report with 0% coverage.

> Decide if this change is breaking - what if someone made the coverage an essential part of their workflow/pipeline (e. g. min cov thresholds)

I would argue that this change is mostly a bugfix. I think it will be rare for pipelines to break on this. And most people who care about coverage, would actually want to have it break on this.  

>  Add documentation

Where should this be added?

wtho:
Ok cool! I'm on board!

This is not a fix though as there was nothing broken before and it is just adding configuration, but I am ok with it going in as a non-breaking feature.

I would like to still have it set for the preset's unit tests as well (in `package.json`), just because it is purely beneficial, not harming at all.

Documentation would be required wherever the configuration in `README.md` is in-place (especially the (exposed-configuration section)[https://github.com/thymikee/jest-preset-angular#exposed-configuration), plus an entry in `CHANGELOG.md`.

@ahnpnl what do you think?
@thymikee can you have a quick look at the discussion?

ahnpnl:
I agreed with @thymikee , the PR only changes the config of the example so it looks fine to me.

wtho:
Actually we discussed if it could go into the preset itself
> Based on this I would propose the following:
> 1. Add this configuration in the preset. Having it there would improve the base configuration of anyone using the preset.

But I think your feedback clarified this proposal as well, we should not add it to the preset itself.

### Some background of this decision
This preset is now used by almost 13k projects on GitHub alone, having over a 100k weekly downloads. Every small change will affect many developers. Making minor changes in the preset will make many developers come back later to us and projects using our preset, opening issues or requiring them to dive into documentations. We only want changes in this preset that actually fix bugs or add non-intrusive, useful features, or breaking features that enable more use cases. 

The option `moduleNameMapper` was added when this preset was created, so it was not breaking for anyone.

-----

You are welcome to contribute by adding this config to the `example` app and the jest config for testing the preset project in `package.json`.

ahnpnl:
The change in this PR looks helpful for a base config, imo it should be added too.

ahnpnl:
close as not applicable anymore

PR #340, opened 2020-01-08
fix: test.each should be in ProxyZone

### LINKED ISSUE #339: Can't run it.each or test.each in fakeAsync zone
I ran into an issue with not being able to run it.each or test.each in a ```fakeAsync``` zone. This sample test
```
  it.each([[1]])('should be equal', fakeAsync((val) => {
    expect(val).toEqual(val);
  }));
``` 
results in the following error
```
  Expected to be running in 'ProxyZone', but it was not found.

      at Function.Object.<anonymous>.ProxyZoneSpec.assertPresent (../node_modules/zone.js/dist/proxy.js:42:19)
```
After some investigation it turns out that describe.each does work so as a test I modified the zone-patch file by adding
```
 const bindTest = originalJestFn =>
  function() {
    const eachArguments = arguments;
    return function(description, specDefinitions, timeout) {
      arguments[1] = wrapTestInZone(specDefinitions);
      return originalJestFn.apply(this, eachArguments).apply(this, arguments);
    };
  };
```
and then changing line 96 from
```
env[methodName].each = originaljestFn.each;
```
to 
```
env[methodName].each = bindTest(originaljestFn.each);
```
This fixed the issue.

I am a little confused as to why the ```describe``` blocks were patched but the other blocks weren't. Was there a reason?
I could submit a pull request for this once I understand if this is the correct fix.

JiaLiPassion:
Close #339.

Fix `test.each` not in ProxyZone issue.
Also remove some not needed assignment.

And I will try to move this source into `angular/zone.js` repo, and also add support to timer methods such as `jest.fakeTimers/jest.runAllTicks`, after that `jest-preset-angular` can just import `zone-testing` bundle.

@thymikee , please help to check this one, thank you!

### DIFF
--- a/example/src/app/service/hero.service.spec.ts
+++ b/example/src/app/service/hero.service.spec.ts
@@ -1,104 +1,135 @@
-import { inject, TestBed } from '@angular/core/testing'
-import { HttpClientModule, HttpErrorResponse, HttpRequest } from '@angular/common/http'
-import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing'
+import {
+  HttpClientModule,
+  HttpErrorResponse,
+  HttpRequest
+} from "@angular/common/http";
+import {
+  HttpClientTestingModule,
+  HttpTestingController
+} from "@angular/common/http/testing";
+import { inject, TestBed } from "@angular/core/testing";
 
-import { heroesUrl, HeroService } from './hero.service'
+import { heroesUrl, HeroService } from "./hero.service";
 
-describe('Service: HeroService', () => {
-  let service: HeroService
-  let backend: HttpTestingController
+describe("Service: HeroService", () => {
+  let service: HeroService;
+  let backend: HttpTestingController;
 
-  const expectedData = {
-    id: 1,
-    name: 'Test hero',
-  }
+  const expectedData = { id: 1, name: "Test hero" };
 
   const expectedDataAll = [
-    {
-      id: 1,
-      name: 'Test hero 1'
-    },
-    {
-      id: 2,
-      name: 'Test hero 2'
-    }
-  ]
+    { id: 1, name: "Test hero 1" },
+    { id: 2, name: "Test hero 2" }
+  ];
 
   beforeEach(() => {
     TestBed.configureTestingModule({
-      imports: [
-        HttpClientModule,
-        HttpClientTestingModule,
-      ],
-      providers: [
-        HeroService,
-      ],
-    })
+      imports: [HttpClientModule, HttpClientTestingModule],
+      providers: [HeroService]
+    });
 
-    backend = TestBed.get(HttpTestingController)
-    service = TestBed.get(HeroService)
+    backend = TestBed.get(HttpTestingController);
+    service = TestBed.get(HeroService);
 
     // Mock implementation of console.error to
     // return undefined to stop printing out to console log during test
-    jest.spyOn(console, 'error').mockImplementation(() => undefined)
-  })
+    jest.spyOn(console, "error").mockImplementation(() => undefined);
+  });
 
-  afterEach(inject([ HttpTestingController ], (_backend: HttpTestingController) => {
-    _backend.verify()
-  }))
+  afterEach(inject(
+    [HttpTestingController],
+    (_backend: HttpTestingController) => {
+      _backend.verify();
+    }
+  ));
 
-  it('should create an instance successfully', () => {
-    expect(service).toBeDefined()
-  })
+  it("should create an instance successfully", () => {
+    expect(service).toBeDefined();
+  });
 
-  it('should call the GET heroes api and return all results', () => {
-    let actualDataAll = {}
+  it("should call the GET heroes api and return all results", () => {
+    let actualDataAll = {};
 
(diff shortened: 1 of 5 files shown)


wtho:
Nice!
Looking forward to the `zone-testing` bundle getting more features!

Could you add a minimal test using `test.each` in the example-app (in this repo at `/example`)? That would be awesome, so we can also track the migration to `zone-testing` later on.

@thymikee you definitely have to have a look at this, as you authored the patch for `zone.js` back then.

-----
Do you plan to migrate all patches from `zone-patch.js` to `zone-testing`, or would that create conflicts with karma?

JiaLiPassion:
@wtho, thanks for the review, sure I will add test under `/example`, and I have tested in another standalone app.
And I am working on the `zone.js` integration with `jest` in `angular` repo, after that, `jest-preset-angular` can just import `zone-testing` and care about the `jest` configuration only.

And thanks for the awesome work for `patch zone` for jest.

JiaLiPassion:
@thymikee, @wtho , I have added test cases under `example`, please review, thanks!

src/zone-patch/index.js
```diff
@@ -48,9 +48,9 @@ function wrapTestInZone(testBody) {
   if (testBody === undefined) {
     return;
   }
-  return testBody.length === 0
-    ? () => testProxyZone.run(testBody, null)
-    : done => testProxyZone.run(testBody, null, [done]);
+  return function() {
+    return testProxyZone.run(testBody, null, arguments);
```
thymikee: How about return `(...args) => testProxyZone.run(testBody, null, args)`? Not that it matters too much, it's just my preference of not using `arguments` when possible

example/src/app/service/hero.service.spec.ts
```diff
@@ -1,104 +1,109 @@
...
-        && req.method === 'GET'
-        && req.params.get('id') === '1'
-    }, `GET hero data from ${heroesUrl}?id=1`)
-      .flush(expectedData)
+  test.each([
+    [1, { id: 1, name: 'Test Hero 1' }],
+    [2, { id: 2, name: 'Test Hero 2' }]
+  ])('should call the GET hero api and return the result', (id: number, testData: any) => {
+    debugger;
+    console.log('id, testData', id, testData, (window as any).Zone.current.name);
```
thymikee: please remove debug code

src/zone-patch/index.js
JiaLiPassion: Sure, I also think that is a good idea, will update it.

example/src/app/service/hero.service.spec.ts
JiaLiPassion: Sorry, forget to remove

JiaLiPassion:
@thymikee , I have updated the PR, also add test cases for `describe.each`, I already test all `jest apis` in another standalone apps. 
Thanks!

example/src/app/simple/simple.component.spec.ts
```diff
@@ -36,3 +35,32 @@ describe('SimpleComponent', () => {
     expect(fixture.nativeElement).toMatchSnapshot();
   });
 });
+
+@Component({
+  selector: 'another-comp',
+  template: ''
+})
+class AnotherComponent {}
```
wtho: Can we move this to another file? Until now we kept one component per file, like Angular recommends.

src/zone-patch/index.js
```diff
@@ -48,61 +48,68 @@ function wrapTestInZone(testBody) {
...
+  env[methodName] = function(...args) {
+    args[1] = wrapDescribeInZone(args[1]);
+    return originaljestFn.apply(this, args);
   };
   env[methodName].each = bindDescribe(originaljestFn.each);
   if (methodName === 'describe') {
     env[methodName].only = env['fdescribe'];
     env[methodName].skip = env['xdescribe'];
-    env[methodName].only.each = bindDescribe(originaljestFn.only.each);
-    env[methodName].skip.each = bindDescribe(originaljestFn.skip.each);
```
wtho: Are the `only.each` and `skip.each` not required anymore? I tested it and it worked, but just to get an insight from you side.

example/src/app/simple/simple.component.spec.ts
JiaLiPassion: Sure, I moved it to another file

src/zone-patch/index.js
JiaLiPassion: Sure, because here https://github.com/thymikee/jest-preset-angular/pull/340/files/bf1a11ee365a9431318d6e3c960b43745b767677#diff-561fc1ab4bca44fdd10cdb2851f0b291R86-R87

`env[methodName].only = env['fit']` and `env[methodName].skip = env['xit']`, and `fit.each` and `xit.each` has been patched, so we don't need to patch `only.each` and `skip.each` again.

wtho: Perfect! Thanks!

stefanmatar:
Can we merge this?

mehrad-rafigh:
Hey guys, any news on this? We are eagerly waiting :)

wtho:
@mehrad-rafigh Sorry for the delay, it's out now.

PR #343, opened 2020-01-27
Merge in class filtering solution from issue 336

### LINKED ISSUE #336: Remove more internal angular attributes from snapshots
Currently, snapshots often contain internal angular attribute values that cause tests to fail. 
PR https://github.com/thymikee/jest-preset-angular/pull/97 fixed some of these but there are still more.  The remaining ones were [discussed in that PR](https://github.com/thymikee/jest-preset-angular/pull/97#issuecomment-485243845) but were shelved at that time.  This new issue tracks the need to continue where that PR left off and remove the additional attribute + attribute values.

This issue is especially problematic when working with Angular Material.  The snapshots will often contain one of the following attribute+values where the numeric indexes change.
* `class=ng-tns-c25-1`
* `aria-owns=mat-input-4`
* `aria-labelledby=mat-input-4`
* `aria-controls=cdk-step-content-2-0`
* `id=cdk-step-content-0-0`
 * `for=mat-form-field-label-9`

Interestingly, the snapshots don't error on repeated executions of the same tests since the numeric indexes remain constant.  However, if you rearrange the order of the same tests or add new tests before existing tests, the indexes change and snapshots fail.  It appears Angular globally stores an index counter (called a `namespaceId`) and sequentially increments them for each render.  If we had a way to reset this global state before each test, then this issue wouldn't be a problem since the indexes would be consistent.  I was not able to find a way to do this albeit [multiple attempts](https://github.com/angular/angular/issues/10984).

A quick workaround is to add a local Serializer and reference it in `snapshotSerializers`.  Source code below:

<details>
<summary>AngularStripAnimationAttributesSnapshotSerializer.ts</summary>

```ts
'use strict';

const jestDOMElementSerializer = require('pretty-format').plugins.DOMElement;

const attributesToClean: { [key: string]: RegExp[] } = {
  class: [/^.*-\w*\d+-\d+$/], // e.g. "ng-tns-c25-1"
  id: [/^.*-\d+$/], // e.g. "mat-input-4", "cdk-step-content-0-0"
  for: [/^.*-\d+$/], // e.g. "mat-form-field-label-9"
  'aria-owns': [/^.*-\d+$/], // e.g. "mat-input-4"
  'aria-labelledby': [/^.*-\d+$/], // e.g. "mat-input-4", "cdk-step-label-0-0"
  'aria-controls': [/^.*-\d+$/], // e.g. "cdk-step-content-2-0"
};

const hasAttributesToClean = (attribute: Attr) => {
  return Object.keys(attributesToClean).some(
    removePatternKey => attribute.name === removePatternKey,
  );
};

const serialize = (node: Element, ...rest: any[]) => {
  const nodeCopy = node.cloneNode(true) as Element;
  Object.values(nodeCopy.attributes)
    .filter(hasAttributesToClean)
    .forEach((attribute: Attr) => {
      attribute.value = attribute.value
        .split(' ')
        .filter((attrValue: string) => {
          return !attributesToClean[attribute.name].some(attributeCleanRegex =>
            attributeCleanRegex.test(attrValue),
          );
        })
        .join(' ');
      nodeCopy.attributes.setNamedItem(attribute);
    });

  return jestDOMElementSerializer.serialize(nodeCopy, ...rest);
};

const serializeTestFn = (val: Element) => {
  return (
    val.attributes !== undefined &&
    Object.values(val.attributes).some(hasAttributesToClean)
  );
};
const test1 = (val: Element) =>
  jestDOMElementSerializer.test(val) && serializeTestFn(val);

module.exports = {
  serialize: serialize,
  test: test1,
};
```
</details>

emanguy:
# Summary

Adds in filtering for contents of attributes as proposed in #336 (thanks for the sample code @ravishivt !)

Fixes #336 

## Test Plan

Tested by adding into storyshots tests for my company, but additional automated testing would be great. Please advise on what I should do for that. Sample from one of the tests:

image

Update: the example app snapshot test verifies this functionality.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

(assuming the other serializer code works, this should too)

## Checklist

- [ ] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I mentioned this change in `CHANGELOG.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/example/src/app/__snapshots__/app.component.spec.ts.snap
+++ b/example/src/app/__snapshots__/app.component.spec.ts.snap
@@ -13,7 +13,6 @@ exports[`AppComponent snaps 1`] = `
       class="custom-class25-0 margin-rem3-0"
     />
     <span>
-
       aaa $1234
     </span>
     <span
--- a/example/src/app/__snapshots__/app.component.spec.ts.snap
+++ b/example/src/app/__snapshots__/app.component.spec.ts.snap
@@ -7,20 +7,20 @@ exports[`AppComponent snaps 1`] = `
   variableWithPrecedingDolar={[Function Number]}
 >
   <h1
-    class="ng-tns-c1-0"
+    class=""
   >
      app works! 
     <app-calc
-      class="ng-tns-c1-0"
+      class=""
     />
     <span
-      class="ng-tns-c1-0"
+      class=""
     >
       aaa $1234
     </span>
     
     <span
-      class="ng-tns-c1-0 ng-star-inserted"
+      class=""
       style=""
     >
       ddd
--- a/example/src/app/__snapshots__/app.component.spec.ts.snap
+++ b/example/src/app/__snapshots__/app.component.spec.ts.snap
@@ -6,21 +6,16 @@ exports[`AppComponent snaps 1`] = `
   title={[Function String]}
   variableWithPrecedingDolar={[Function Number]}
 >
-  <h1
-    class=""
-  >
+  <h1>
      app works! 
     <app-calc
-      class=""
+      class="custom-class25-0 margin-rem3-0"
     />
-    <span
-      class=""
-    >
+    <span>
       aaa $1234
     </span>
     
     <span
-      class=""
       style=""
     >
       ddd
--- a/src/AngularNoNgAttributesSnapshotSerializer.js
+++ b/src/AngularNoNgAttributesSnapshotSerializer.js
@@ -39,11 +39,11 @@ const serialize = (node, ...rest) => {
           );
         })
         .join(' ');
-	  if (attribute.value === '') {
+      if (attribute.value === '') {
         nodeCopy.attributes.removeNamedItem(attribute.name);
-	  } else {
+      } else {
         nodeCopy.attributes.setNamedItem(attribute);
-	  }
+      }
     });
 
   return jestDOMElementSerializer.serialize(nodeCopy, ...rest);
--- a/src/AngularNoNgAttributesSnapshotSerializer.js
+++ b/src/AngularNoNgAttributesSnapshotSerializer.js
@@ -4,12 +4,12 @@ const jestDOMElementSerializer = require('pretty-format').plugins.DOMElement;
 
 const attributesToRemovePatterns = ['ng-reflect', '_nghost', '_ngcontent', 'ng-version'];
 const attributesToClean = {
-  class: [/^.*-\w*\d+-\d+$/, /^ng-star-inserted$/], // e.g. "ng-tns-c25-1" or "ng-star-inserted", literally
-  id: [/^.*-\d+$/], // e.g. "mat-input-4", "cdk-step-content-0-0"
-  for: [/^.*-\d+$/], // e.g. "mat-form-field-label-9"
-  'aria-owns': [/^.*-\d+$/], // e.g. "mat-input-4"
-  'aria-labelledby': [/^.*-\d+$/], // e.g. "mat-input-4", "cdk-step-label-0-0"
-  'aria-controls': [/^.*-\d+$/], // e.g. "cdk-step-content-2-0"
+  class: [/^(?:mat|cdk|ng).*-\w*\d+-\d+$/, /^ng-star-inserted$/], // e.g. "ng-tns-c25-1" or "ng-star-inserted", literally
+  id: [/^(?:mat|cdk|ng).*-\d+$/], // e.g. "mat-input-4", "cdk-step-content-0-0"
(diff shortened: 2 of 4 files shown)


emanguy:
The test above failed because the no NG attributes serializer removed the auto-added angular classes as expected. Don't know if that helps the case for the fact that this works.

ahnpnl:
- I think you can update snapshot of the failed test.
- Is it possible to remove `ng-star-inserted` in this PR too ?

emanguy:
How do I update the snapshot? I didn't see anything referencing the snapshots in the npm scripts

ahnpnl:
Hi, you can do this:
- cd examples
- yarn test -u
Indeed there is no update snapshot script but you just need to use jest cli option.

emanguy:
Ah, I didn't realize there was a separate NPM project in the example directory. I've made the changes you requested.

src/AngularNoNgAttributesSnapshotSerializer.js
```diff
@@ -3,23 +3,51 @@
...
 
 const attributesToRemovePatterns = ['ng-reflect', '_nghost', '_ngcontent', 'ng-version'];
+const attributesToClean = {
+  class: [/^.*-\w*\d+-\d+$/, /^ng-star-inserted$/], // e.g. "ng-tns-c25-1" or "ng-star-inserted", literally
+  id: [/^.*-\d+$/], // e.g. "mat-input-4", "cdk-step-content-0-0"
+  for: [/^.*-\d+$/], // e.g. "mat-form-field-label-9"
+  'aria-owns': [/^.*-\d+$/], // e.g. "mat-input-4"
+  'aria-labelledby': [/^.*-\d+$/], // e.g. "mat-input-4", "cdk-step-label-0-0"
+  'aria-controls': [/^.*-\d+$/], // e.g. "cdk-step-content-2-0"
+};
```
wtho: All of the classes do start with `mat`, `cdk` or `ng` and I think we really should take advantage of this.
There might some project using `component-nr25-1` and their class would "randomly" be removed.
As this is a unit test library that usually does not care about styles, I can understand arguments that the "damage" is negligible, but I think it's too aggressive to just remove all classes that end on the pattern.

src/AngularNoNgAttributesSnapshotSerializer.js
```diff
@@ -3,23 +3,51 @@
...
+    .forEach(attribute => {
+      attribute.value = attribute.value
+        .split(' ')
+        .filter(attrValue => {
+          return !attributesToClean[attribute.name].some(attributeCleanRegex => 
+              attributeCleanRegex.test(attrValue),
+          );
+        })
+        .join(' ');
+      nodeCopy.attributes.setNamedItem(attribute);
```
wtho: We can probably remove the attribute completely if `attribute` is empty, so we avoid `class=""` in the template.

example/src/app/__snapshots__/app.component.spec.ts.snap
```diff
@@ -7,20 +7,20 @@ exports[`AppComponent snaps 1`] = `
   variableWithPrecedingDolar={[Function Number]}
 >
   <h1
-    class="ng-tns-c1-0"
+    class=""
   >
```
wtho: Can you add a class value where a in-template-defined class remains in the snapshot? Maybe also some edge case like `custom-class25-0`.
A valid use case I can come up with right now is using classes for special padding or margin values like `margin-rem3-0`.

emanguy:
Sorry this has gotten stale. I'll get around to making the changes this weekend.

emanguy:
Alright! Changes are made. The regexes are now more targeted at things following those prefixes.

ahnpnl:
LGTM, only needs to resolve the conflict and add `CHANGELOG.md` 👍 

Probably this also needs to be ivy compatible, but that will be for another PR imo.

intellix:
How about also this which creeped in since updating to 8.1.3 with ivy support:
```
  __ngContext__={[Function LRootView]}
```

Also, you said ivy compatible can come after but that support is already in master so... required?

ahnpnl:
The current snapshot serializer in 8.1.3 is already compatible with ivy. This PR is only an enhancement to get rid of unwanted attributes :)

emanguy:
Changelog is updated, I'm hoping you intended for me to put it in the 8.1.3 changelog!

Merge conflict is also resolved.

wtho:
No, 8.1.3 is already released. I will fix in on the 8.2 release though.
Thanks a lot!

wtho:
Released in v8.2.0.

PR #346, opened 2020-02-06
Jest 25

### LINKED ISSUE #325: Remove jsdom v15 environment once jest v25 is out
Jest v25 does use JSDOM v15 [by default](https://github.com/facebook/jest/pull/8851), so we do not have to define this version in the preset anymore.

This issue is a reminder and can be solved by a simple PR once jest v25 is out.

### LINKED ISSUE #344: Compatibility with Jest 25
```
warning " > jest-preset-angular@8.0.0" has incorrect peer dependency "jest@^24.1.0".
warning "jest-preset-angular > ts-jest@24.0.2" has incorrect peer dependency "jest@>=24 <25".
```

ahnpnl:
# Summary
Jest 25 compatible.
Closes #344 
Closes #325 

## Test Plan
- CI green
- No dependency warning

## Checklist

- [ ] I have tested this on a device and a simulator
- [x] I added the documentation in `README.md`
- [x] I mentioned this change in `CHANGELOG.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)


README.md
```diff
@@ -476,7 +475,7 @@ Reference: https://github.com/facebook/jest/issues/708
 By default we use JSDOM v15, which requires Node v8+. If you want to use Node in a lower version to run your tests, make sure to configure jest to use an older JSDOM version, either in your `jest.config.js`, `jest.config.json` or `package.json`:
 
 ```json
-"testEnvironment": "jsdom"
+"testEnvironment": <jsdom_environment_package_<_15>
 ```
 
 If you use JSDOM v11 or lower, you might have to mock `localStorage` or `sessionStorage` on your own or using some third-party library by loading it in `setupFilesAfterEnv`.
```
thymikee: Can we remove paragraphs about JSDOM that are no longer necessary? Jest comes with JSDOM v15 now. If we want to guide users to use newer one, let's update the messaging.

ahnpnl: hmm I think indeed the paragraph is not necessary anymore. What I intended was to provide a fallback user guide but do we need a fallback anymore ? If we don't need, we can just remove the entire paragraph.

thymikee: On the other hand there's jsdom@16 that Jest 25 can't use, because it dropped Node 8. So the small reference can still be helpful to guide users too. But shouldn't be too lengthy

ahnpnl: I propose:

#### Configure other JSDOM versions

If you want to use another JSDOM version, either in your jest.config.js, jest.config.json or package.json:

```
"testEnvironment": <jsdom_environment_package>
```

thymikee: That's too vague, people will be confused. Let's use concrete example: https://www.npmjs.com/package/jest-environment-jsdom-sixteen

wtho: Yeah, referencing the current version seems a good approach here.
Btw, if you could also remove the `testEnvironment` line from the exposed config in `README.md` would be awesome!

ahnpnl: I think it has been removed already :)

README.md
```diff
@@ -471,12 +470,16 @@ The same declaration can be applied to other vendor libraries.
...
 
-### Failing with Node v7 or lower
+### Configure other JSDOM versions
 
-By default we use JSDOM v15, which requires Node v8+. If you want to use Node in a lower version to run your tests, make sure to configure jest to use an older JSDOM version, either in your `jest.config.js`, `jest.config.json` or `package.json`:
+**Jest** v25 by default uses **JSDOM** 15 to support Node 8, which should also be compatible with Jest 24 and earlier.
 
-```json
-"testEnvironment": "jsdom"
+If you need a newer JSDOM than the one that ships with Jest, install this package using npm install --save-dev jest-environment-jsdom-sixteen or yarn add jest-environment-jsdom-sixteen --dev, and edit your Jest config like so:
```
thymikee: Feel free to send a followup :) 
```suggestion
If you need a newer JSDOM than the one that ships with Jest and run Node >=10, you can install `jest-environment-jsdom-sixteen` package, and edit your Jest config like so:
```

README.md
wtho: Oh yeah, sorry about that!

PR #361, opened 2020-03-13
fix: refactor zone patch test method

### LINKED ISSUE #359: regression in zone patch in 8.1.1 with it.each as tagged template literal
The fix released in 8.1.1 didn't consider usage of it.each as tagged template literal and currently it breaks with 
```
 FAIL  src/app/app.component.spec.ts (12.248s)
  ● AppComponent › encountered a declaration exception

    TypeError: row.map is not a function

      113 |     foo  | bar
      114 |     ${1} | ${2}
    > 115 |   `('it.each should work with table as a tagged template literal', ({foo, bar}) => {
          |    ^
      116 |     expect(foo).toBe(1);
      117 |     expect(bar).toBe(2);
      118 |   });

      at node_modules/jest-preset-angular/build/zone-patch/index.js:137:73
          at Array.map (<anonymous>)
      at src/app/app.component.spec.ts:115:4
      at Object.<anonymous> (src/app/app.component.spec.ts:9:1)
```
This usage works in 8.1.0.  ~~I'll submit a PR with regression test in example app~~ it's here: #360

JiaLiPassion:
refactor the zone.js patch test method logic to handle all cases
1. each with args
2. each with args and done
3. each with template string
4. each with template string and done

Close #359.

### DIFF
--- a/src/zone-patch/index.js
+++ b/src/zone-patch/index.js
@@ -44,51 +44,27 @@ function wrapDescribeInZone(describeBody) {
 // Create a proxy zone in which to run `test` blocks so that the tests function
 // can retroactively install different zones.
 const testProxyZone = ambientZone.fork(new ProxyZoneSpec());
-function wrapTestInZone(testBody, eachArgs) {
+function wrapTestInZone(testBody) {
   if (testBody === undefined) {
     return;
   }
 
-  if (!eachArgs || eachArgs.length === 0 || eachArgs[0].length === 0) {
-    // If we are not handling `test.each`, then the parameter of `testBody`
-    // will be 0 or 1, if it is 1, then we need to return a function with
-    // done parameter
+  const wrappedFunc = function() {
+    return testProxyZone.run(testBody, null, arguments);
+  };
+  try {
+    Object.defineProperty(wrappedFunc, 'length', {
+      configurable: true,
+      writable: true,
+      enumerable: false
+    });
+    wrappedFunc.length = testBody.length;
+  } catch (e) {
     return testBody.length === 0
       ? () => testProxyZone.run(testBody, null)
       : done => testProxyZone.run(testBody, null, [done]);
-  } else {
-    // Dynamically create a Function to contain the same length
-    // of the parameters as the testBody
-    // For example:
-    // ```
-    // test.each([[1, 2]])('test.each', (arg1, arg2) => {});
-    // ```
-    // In this case we need to return a function like this
-    // ```
-    // return function(arg1, arg2) {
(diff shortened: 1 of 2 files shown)


src/zone-patch/index.js
```diff
@@ -44,51 +44,27 @@ function wrapDescribeInZone(describeBody) {
...
-    }
-    args.push('testBody');
-    args.push('testProxyZone');
-    if (len < testBody.length) {
-      args.push('done');
-      argString += ', done';
-    }
-    const funcBody = `
-      return testProxyZone.run(testBody, null, [${argString}])`;
-    return new Function(args, funcBody);
```
wtho: Is the previous fix not required anymore?

src/zone-patch/index.js
```diff
@@ -108,29 +84,7 @@ const bindDescribe = originalJestFn =>
...
-      args[1] = wrapTestInZone(args[1], ...eachArgs);
-      if (testBody.length > 0 || (eachArgs.length > 0 && eachArgs[0].length > 0)) {
-        eachArgs.forEach(row => {
-          const modifiedRow = row.map(a => {
-            a.push(testBody);
-            a.push(testProxyZone);
-          });
-          return modifiedRow;
-        });
-      }
```
wtho: Can you elaborate how this is all covered in the new fix?

src/zone-patch/index.js
JiaLiPassion: @wtho , yeah, the reason the previous version is to make sure to return a function with the same length as the test function.
For example,

```
test.each([[1,2]])((arg1, arg2) => {
  expect(arg1).toBe(1);
  expect(arg2).toBe(2);
})
```

So the test function is  `(arg1, arg2) => {...}`, and we need to wrap the function and return a new function to let it run in the `ProxyZone`.
But if we return the function like this
```
return function() {
  return originalTestBody.apply(this, arguments);
}
```
the returned `function.length` will be `0`, not `2` in this example, and `jest` need the function.length to check whether it has `done()` or not.

So in the previous version, I returned a dynamic function with the correct length.
But the previous commit didn't consider the `template literal case`, so the bug occurs.

And in fact, I don't need such complex logic, what I need is just return a function with dynamic length.

```
const wrappedFunc = function() {return originalFunc.apply(this, arguments);
wrappedFunc.length = originalFunc.length;
return wrappedFunc;
``` 

Will handle all the cases.

src/zone-patch/index.js
JiaLiPassion: The same reason with the explanation above. Thanks.

PR #374, opened 2020-05-16
feat: support jest 26

### LINKED ISSUE #367: Support Jest 26
## 🚀 Feature Proposal

Jest 26 has released. We want to add support for Jest 26 once ts-jest releases a compatible version.

## Motivation

Make developers feel comfortable to work

## TODO

- Upgrade ts-jest and pretty-format dependencies to ^26.0.0

- Set jest peerDependency to ^26.0.0

- Upgrade jest to 26.0.0 and typescript to 3.8.3

- Upgrade dependencies for example project.

- Update README regarding to jest-environment-jsdom 16 which is default in jest 26

Since jest 26 requires TypeScript 3.8.3, default jsdom 16 this support should be in a major version.

ahnpnl:
Close #367


README.md
```diff
@@ -479,15 +479,7 @@ Reference: https://github.com/facebook/jest/issues/708
...
-**Jest** v25 by default uses **JSDOM** 15 to support Node 8, which should also be compatible with Jest 24 and earlier.
-
-If you need a newer JSDOM than the one that ships with Jest and run Node >=10, you can install `jest-environment-jsdom-sixteen` package, and edit your Jest config like so:
-
-```
-{
-  "testEnvironment": "jest-environment-jsdom-sixteen"
-}
-```
+**Jest** v26 by default uses **JSDOM** 16 to support Node 10+.
```
ahnpnl, line 482: Is JSDOM 16 compatible with Jest 25 ? I'm not sure what to write here. Any thoughts ?

wtho: It should be. But in 99% of the cases, whoever uses Jest v26 will be happy to use JSDOMv16 and Nodev10.
JSDOM is not a interesting test-target like IE11 or Edge, so a specific setup will not be wanted by 99% of jest users.

I think we can remove that "should also be compatible" part like you did 👍

thymikee: `jest-environment-jsdom-sixteen ` et. al. are compatible with any version of Jest. Knowing jsdom project, they'll drop Node 10 way earlier than Jest does, so this section may still be handy. You could provide references to specific tags, so people on lower versions can refer to

ahnpnl: I brought back the part for providing references to specific tags as suggestion :)

ahnpnl:
shall I merge this ?

thymikee:
Let's do it

dzonatan:
When to expect a release?

cedricduffournet:
Can this PR be added to a new release ?

wtho:
Was released in 8.2.1.

PR #384, opened 2020-05-21
Add Dependabot badge

### LINKED ISSUE #383: Add dependabot badge to README.md
This bade shows that dependabot is enabled in our repository: https://badgen.net/dependabot/thymikee/jest-preset-angular?icon=dependabot

There is also this a badge for the semver-score, but I think this is not so relevant for us as we are only a dev dependency to our projects:
https://dependabot.com/compatibility-score.html/?dependency-name=puma&package-manager=bundler&version-scheme=semver

AlQuraian:
Resolves #383

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,5 +1,0 @@
-version: 1
-update_configs:
-  - package_manager: "javascript"
-    directory: "/"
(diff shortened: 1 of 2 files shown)


ahnpnl:
Will you please also remove greenkeeper badge ?

PR #406, opened 2020-07-07
feat: angular ctor params transformer

### LINKED ISSUE #288: Angular 8: Can't resolve all parameters for Component: (?).

wtho:
Introduces the usage of the @ngtools/webpack ctor-params transformer to
downlevel angular decorators to static properties

This also is the first usage of an Angular transformer in our setup. By using the compilation tools provided by the Angular Team to compile the code, we ensure the tests treat the code like production code. Furthermore this facilitates keeping up with the Angular Team.

### Checklist
- [X] Transformer using Angular's ctor-params
- [x] documentation in `README.md` regarding `isolatedModules: true`
- [x] update `CHANGELOG.md`
- [ ] Test for the transformer

Closes #288.

### Further Discussion
While we are finally able to include Angular transformers, and in this case to circumvent the issue with emitting more metadata than Angular itself, we are still behind Angular Ivy compilation in jest + angular. This could change if we take a better look at the compilation behavior of the [AngularCompilerPlugin](https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/angular_compiler_plugin.ts) as well as the used [transformers](https://github.com/angular/angular-cli/tree/master/packages/ngtools/webpack/src/transformers). While partial functionality is definitely not interesting in test scenarios, other features are necessary. There might also be more compiling functionality in `@angular/compiler-cli`. More research has to be done in this field.

### Note for Compatibility with Angular v10
In Angular v10 the `ctor-parameters.js` disappeared (in [this commit](https://github.com/angular/angular-cli/commit/0a0be3b37b2daa8b77887e3873050f667fef4ce1)). The reference in the `AngularConstructorParameters` has to be updated to be compatible with Angular v10 (and therefore incompatible with Angular v9). The current transformer handles both scenarios by trying to load `ctor-parameters` first and then falling back to the newer `compiler-cli` transformer.

### DIFF
--- a/example/jest.config.js
+++ b/example/jest.config.js
@@ -1,10 +1,5 @@
 module.exports = {
   preset: "jest-preset-angular",
-  snapshotSerializers: [
-    "jest-preset-angular/build/AngularNoNgAttributesSnapshotSerializer.js",
-    "jest-preset-angular/build/AngularSnapshotSerializer.js",
-    "jest-preset-angular/build/HTMLCommentSerializer.js"
-  ],
   moduleNameMapper: {
     "\\.(jpg|jpeg|png)$": "<rootDir>/__mocks__/image.js",
     "^@lib/(.*)$": "<rootDir>/src/lib/$1"
--- a/src/AngularConstructorParametersTransformer.ts
+++ b/src/AngularConstructorParametersTransformer.ts
@@ -1,0 +1,71 @@
+import type {
+  TransformerFactory,
+  SourceFile,
+  TransformationContext,
+  Program,
+} from "typescript";
+import { ConfigSet } from "ts-jest/dist/config/config-set";
+
+function tryLoadV8V9Transformer(
+  program: Program
+): TransformerFactory<SourceFile> | undefined {
+  try {
+    const {
+      downlevelConstructorParameters,
+    } = require("@ngtools/webpack/src/transformers/ctor-parameters");
+
+    return downlevelConstructorParameters(() => program.getTypeChecker());
+  } catch (err) {
+    return undefined;
+  }
+}
+
+function tryLoadV10Transformer(
+  program: Program
+): TransformerFactory<SourceFile> | undefined {
+  try {
+    const {
+      constructorParametersDownlevelTransform,
+    } = require("@angular/compiler-cli/src/tooling");
+
+    return (context: TransformationContext) => {
+      const compilerCliFactory = constructorParametersDownlevelTransform(
+        program
+      );
+      return compilerCliFactory(context);
+    };
+  } catch (err) {
+    return undefined;
+  }
+}
+
+/**
+ * The factory of hoisting transformer factory
+ * @internal
+ */
+export function factory(configSet: ConfigSet): TransformerFactory<SourceFile> {
+  const { program } = configSet.tsCompiler;
+  if (!program) {
+    throw new Error(
+      "TypeScript Compiler module is not provided! Please do not use the `AngularConstructorParametersTransformer` with `isolatedModules: true`"
+    );
+  }
+
+  // Depending on the used Angular version, we have to load the constructor parameters transformer from different locations
+
+  // Angular v8/v9 downlevel constructor transformer
+  let transformerFactory = tryLoadV8V9Transformer(program);
+
+  // Angular v10+ downlevel constructor transformer
+  if (transformerFactory === undefined) {
+    transformerFactory = tryLoadV10Transformer(program);
+  }
+
+  if (transformerFactory === undefined) {
+    throw new Error(
+      "Could not load Constructor Parameters Downlevel Transformer!"
+    );
+  }
+
+  return transformerFactory;
+}
--- a/src/InlineFilesTransformer.ts
+++ b/src/InlineFilesTransformer.ts
@@ -23,7 +23,7 @@
  */
 
 // only import types, for the rest use injected `ConfigSet.compilerModule`
-import {
+import type {
   Node,
   SourceFile,
   TransformationContext,
@@ -32,7 +32,11 @@ import {
   PropertyAssignment,
   Identifier
 } from 'typescript';
-import { getCreateStringLiteral, ConfigSet } from './TransformUtils';
+import type * as _ts from 'typescript';
+import type { ConfigSet } from 'ts-jest/dist/config/config-set';
+import { getCreateStringLiteral } from './TransformUtils';
+
+type TTypeScript = typeof _ts;
 
 // replace original ts-jest ConfigSet with this simple interface, as it would require
 // jest-preset-angular to add several babel devDependencies to get the other types
@@ -53,23 +57,10 @@ const REQUIRE = 'require';
 const TRANSFORM_PROPS = [TEMPLATE_URL, STYLE_URLS];
 
 /**
- * Transformer ID
- * @internal
- */
-export const name = 'angular-component-inline-files';
-
-// increment this each time the code is modified
-/**
- * Transformer Version
- * @internal
- */
-export const version = 1;
-
-/**
  * The factory of hoisting transformer factory
  * @internal
  */
-export function factory(cs: ConfigSet) {
+export function factory(cs: ConfigSet & { compilerModule: TTypeScript }) {
   /**
    * Our compiler (typescript, or a module with typescript-like interface)
    */
@@ -95,7 +86,7 @@ export function factory(cs: ConfigSet) {
    * Clones the assignment and manipulates it depending on its name.
    * @param node the property assignment to change
    */
-  function transfromPropertyAssignmentForJest(node: PropertyAssignment) {
+  function transformPropertyAssignmentForJest(node: PropertyAssignment) {
     const mutableAssignment = ts.getMutableClone(node);
 
     const assignmentNameText = (mutableAssignment.name as Identifier).text;
@@ -156,7 +147,7 @@ export function factory(cs: ConfigSet) {
       // this is an assignment which we want to transform
       if (isPropertyAssignmentToTransform(node)) {
         // get transformed node with changed properties
-        resultNode = transfromPropertyAssignmentForJest(node);
+        resultNode = transformPropertyAssignmentForJest(node);
       }
 
       // look for interesting assignments inside this node in any case
--- a/src/StripStylesTransformer.ts
+++ b/src/StripStylesTransformer.ts
@@ -22,7 +22,7 @@
  */
 
 // only import types, for the rest use injected `ConfigSet.compilerModule`
-import {
+import type {
   Node,
   SourceFile,
   TransformationContext,
@@ -34,7 +34,10 @@ import {
   CallExpression,
   ObjectLiteralExpression
 } from 'typescript';
-import { ConfigSet } from './TransformUtils';
+import type * as _ts from 'typescript';
+import type { ConfigSet } from 'ts-jest/dist/config/config-set';
+
+type TTypeScript = typeof _ts;
 
 /** Angular component decorator Styles property name */
 const STYLES = 'styles';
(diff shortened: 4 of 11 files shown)


src/AngularConstructorParametersTransformer.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The factory of hoisting transformer factory
+ * @internal
+ */
+export function factory(configSet: ConfigSet) {
+  /**
+   * Our compiler (typescript, or a module with typescript-like interface)
+   */
+  const ts = configSet.compilerModule;
```
ahnpnl, line 26: this one looks redundant to me

src/TransformUtils.ts
```diff
@@ -5,6 +5,11 @@ import * as TS from 'typescript';
 // inside the ConfigSet right
 export interface ConfigSet {
   compilerModule: typeof TS;
+  tsCompiler: TsCompiler;
```
ahnpnl, line 8: probably later `ts-jest` can improve `ConfigSet` typings to only expose necessary typings and we won't need this one anymore. Right now it was already improved a lot

```
export declare class ConfigSet {
    readonly parentOptions?: TsJestGlobalOptions | undefined;
    get versions(): Record<string, string>;
    get tsCompiler(): TsCompiler;
    get tsJestDigest(): string;
    readonly logger: Logger;
    constructor(jestConfig: Config.ProjectConfig, parentOptions?: TsJestGlobalOptions | undefined, parentLogger?: Logger);
}
```

src/TransformUtils.ts
```diff
@@ -5,6 +5,11 @@ import * as TS from 'typescript';
 // inside the ConfigSet right
 export interface ConfigSet {
   compilerModule: typeof TS;
```
ahnpnl, line 7: if the `const ts = configSet.compilerModule` above is redundant and can be removed, this `compilerModule` can be removed too. In the latest version of `ts-jest`, this property is no longer exposed as public typings.

src/TransformUtils.ts
wtho: Yeah, that's great! I think I once copied the own version, as it introduced some babel dependencies.

src/TransformUtils.ts
wtho: Do you mean by this that `configSet.compilerModule` equals `configSet.tsCompiler`?

ahnpnl: actually those return 2 different things, `compilerModule` returns the current typescript package of the project while `tsCompiler` returns the compiler instance created internally in `ts-jest`. 

But in `AngularConstructorParametersTransformers.ts`, `compilerModule` isn't used anywhere so I think this property `compilerModule` can be removed.

src/TransformUtils.ts
ahnpnl: maybe using `import type` from TypeScript 3.8 can avoid those ?

src/TransformUtils.ts
wtho: In the other transformers we still need functionality like `configSet.compilerModule.getMutableClone(node)`, which was usually injected by the `compilerModule`, being an instance of `typescript`.

ahnpnl: I see, so I think `ts-jest` should expose that one again in public typings 👍

src/TransformUtils.ts
wtho: Yeah, the babel issue is already resolved, I will reference the ConfigSet from `ts-jest` in all transformers :+1:

src/AngularConstructorParametersTransformer.ts
wtho: It definitely is!

src/TransformUtils.ts
wtho: Yes, this would be great: `compilerModule: typeof TypeScript` where `import * as TypeScript from 'typescript'`

ahnpnl: will be like this

```
export declare class ConfigSet {
    readonly parentOptions?: TsJestGlobalOptions | undefined;
    get versions(): Record<string, string>;
    get compilerModule(): TTypeScript;
    get tsCompiler(): TsCompiler;
    get tsJestDigest(): string;
    readonly logger: Logger;
    constructor(jestConfig: Config.ProjectConfig, parentOptions?: TsJestGlobalOptions | undefined, parentLogger?: Logger);
}
```

where `TTypeScript` is
```
import * as _ts from 'typescript';
export declare type TTypeScript = typeof _ts;
```

It was public before but then marked as internal 😃

wtho: If you assume, that only `ts-jest` authors transformers, it is correctly marked as internal :smile:

wtho: For now I will extend the type with the `compilerModule` inline. Let me know when there's a new version out with the public typings!

src/InlineFilesTransformer.ts
```diff
@@ -32,7 +32,9 @@ import {
   PropertyAssignment,
   Identifier
 } from 'typescript';
-import { getCreateStringLiteral, ConfigSet } from './TransformUtils';
+import * as TS from 'typescript';
+import { ConfigSet } from "ts-jest/dist/config/config-set";
```
ahnpnl, line 36: this can be `import type`

src/StripStylesTransformer.ts
```diff
@@ -34,7 +34,8 @@ import {
   CallExpression,
   ObjectLiteralExpression
 } from 'typescript';
-import { ConfigSet } from './TransformUtils';
+import * as TS from 'typescript';
+import { ConfigSet } from "ts-jest/dist/config/config-set";
```
ahnpnl, line 38: this can be `import type` as well

example/jest.config.js
```diff
@@ -1,10 +1,5 @@
 module.exports = {
   preset: "jest-preset-angular",
-  snapshotSerializers: [
```
ahnpnl, line 3: with ctor params transformer, example app doesn't need these anymore ?

src/AngularConstructorParametersTransformerv10.ts
```diff
@@ -0,0 +1,35 @@
...
+import { TransformationContext } from "typescript";
+// this seems to not be included in a module?
+const { constructorParametersDownlevelTransform } = require('@angular/compiler-cli/src/tooling');
+
+
+/**
+ * Transformer ID
+ * @internal
+ */
+export const name = 'angular-ctor-parameters';
```
ahnpnl, line 11: i don't see this use anywhere. Is it required ?

src/AngularConstructorParametersTransformerv10.ts
```diff
@@ -0,0 +1,35 @@
...
+
+// increment this each time the code is modified
+/**
+ * Transformer Version
+ * @internal
+ */
+export const version = 1;
```
ahnpnl, line 18: same as above

src/AngularConstructorParametersTransformerv8v9.ts
```diff
@@ -0,0 +1,29 @@
+import { ConfigSet } from "ts-jest/dist/config/config-set";
+// this seems to not be included in a module?
+const { downlevelConstructorParameters } = require('@ngtools/webpack/src/transformers/ctor-parameters');
+
+/**
+ * Transformer ID
+ * @internal
+ */
+export const name = 'angular-ctor-parameters';
```
ahnpnl, line 9: same as above

src/AngularConstructorParametersTransformerv8v9.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// increment this each time the code is modified
+/**
+ * Transformer Version
+ * @internal
+ */
+export const version = 1;
```
ahnpnl, line 16: same as above

example/jest.config.js
wtho: We added all three serializers as default preset serializers at some point, so the declaration is redundant.

src/AngularConstructorParametersTransformerv10.ts
wtho: This is just following the convention of the [ts-jest hoist transformer](https://github.com/kulshekhar/ts-jest/blob/master/src/transformers/hoist-jest.ts). If this is unused, I will remove it.

ahnpnl: oh those are for internal `ts-jest` unit test. So indeed they can be removed here :)

wtho: Ok, I thought these might be used by `jest` for some reason. Will remove them.

wtho:
I tried to setup a test, but this requires fiddling a lot with module mocks, specifically to handle the different scenarios of `require(@ngtools/webpack/...)` and `require(@angular/compiler-cli/...)` which is not easy, as the test still tries to load the package, even if `jest.mock('@ngtools/webpack/...', () => ...)` is called. I would need more help with this if you think this is required.

The transformer passes with Angular v9 and v10 projects, so it is sufficient in my opinion.

wtho:
It's ready to merge. Would you like to have a quick look, @thymikee?

ahnpnl:
Ya I think the unit tests can be added later. 

I ran into similar thing when trying to mock typescript. Ended up I used this way Have you tried similar like this https://github.com/ahnpnl/ts-jest/blob/ec15732c29c6cc2767ef10029f417b50990a6b26/src/compiler/language-service.spec.ts#L114

ahnpnl:
what do you think about in the meantime, we can have 1 example app is ng 9, 1 example app is ng 10 to run the tests for this change ?

wtho:
I think that's a good workaround.

But instead of introducing e. g. `/example9/` and `/example10/`, I would prefer to setup a conventional monorepo (and also rename `example` to `test-app`, as there was often confusion, that `example` is the default example of this library):
```
github.com/thymikee/jest-preset-angular
├──jest-preset-angular
├──test-app-v9
└──test-app-v10
```
Or in lerna-fashion:
```
github.com/thymikee/jest-preset-angular
└──packages
    ├──jest-preset-angular
    ├──test-app-v9
    └──test-app-v10
```

What do you think? I'd tackle this in a different PR, and this one would be rebased on top of it.

ahnpnl:
I like the version like `lerna` but since the preset is only on its own but not exposing anything so I'd prefer to go for this

> ```
> github.com/thymikee/jest-preset-angular
> ├──jest-preset-angular
> ├──test-app-v9
> └──test-app-v10
> ```

ya it's wise to do in another PR

wtho:
Ok, so in this case I would prefer the setup
```
github.com/thymikee/jest-preset-angular
├──jest-preset-angular
└──e2e
    ├──test-app-v9
    └──test-app-v10
```
just to avoid unnecessary folder nesting. The container-folder `packages` could be added at any later time, if required one day.

ahnpnl:
ok let's proceed like that

src/TransformUtils.ts
ahnpnl: FYI, `ts-jest` 26.1.2 released. It included the exposed typing for `compileModule`.

src/InlineFilesTransformer.ts
```diff
@@ -52,24 +56,11 @@ const REQUIRE = 'require';
...
- * @internal
- */
-export const version = 1;
-
 /**
  * The factory of hoisting transformer factory
  * @internal
  */
-export function factory(cs: ConfigSet) {
+export function factory(cs: ConfigSet & { compilerModule: TTypeScript }) {
```
ahnpnl, line 63: This factory type isn’t the same as the one in AngularConstructorParametersTransformer.ts. Is it intended ?

src/StripStylesTransformer.ts
```diff
@@ -43,24 +46,11 @@ const COMPONENT = 'Component';
...
- * @internal
- */
-export const version = 1;
-
 /**
  * The factory of hoisting transformer factory
  * @internal
  */
-export function factory(cs: ConfigSet) {
+export function factory(cs: ConfigSet & { compilerModule: TTypeScript }) {
```
ahnpnl, line 53: This factory type isn’t the same as the one in AngularConstructorParametersTransformer.ts. Is it intended ?

ahnpnl:
we might need to postpone this, see my comment https://github.com/thymikee/jest-preset-angular/issues/409#issuecomment-673996044

ahnpnl:
@wtho do you want to include this in 8.3.0 release ?

ahnpnl:
is taken care in #562

PR #410, opened 2020-07-12
build(dev-infra): setup basic e2e structure

ahnpnl:
## Summary

- Added `scripts` folder which includes `e2e.js` to run e2e tests with example apps. This folder also contains some helpers.

- Added an example app using Angular 10 with basic setup for `jest` and `jest-preset-angular`.

- Added new scripts to `package.json`.

## Test plan

- Check out this branch and run `yarn test:e2e` in your local.

## Additional information

This PR only creates a basic e2e structure setup but won't include this into CI yet. The next PR will do it (also including moving the app in `example` to `e2e` folder)


.npmignore
```diff
@@ -1,3 +1,37 @@
...
+src
+
+# Developement scripts
+scripts
+
+# Coverage directory used by tools like istanbul
+coverage
+
+# nyc test coverage
+.nyc_output
```
wtho, line 15: Do we have nyc test coverage?

.npmignore
```diff
@@ -1,3 +1,37 @@
...
+
+# Dependency directories
+node_modules
+jspm_packages
```
wtho, line 19: Do we use `jspm_packages`?

package.json
```diff
@@ -8,13 +8,16 @@
...
   "devDependencies": {
-    "@types/jest": "^26.0.0",
-    "@types/node": "^14.0.4",
-    "jest": "^26.0.1",
+    "@types/jest": "26.x",
+    "@types/node": "12.x",
+    "cross-spawn": "7.x",
+    "fs-extra": "9.x",
+    "jest": "26.x",
+    "lodash.memoize": "4.x",
```
wtho, line 20: Is there any advantage of `26.x` over `^26.0.0`? Usually only the `^`-syntax is used, also by tools like `npm update` or `npm upgrade` or dependabot.

.npmignore
ahnpnl: Oh this we don’t need, copy paste mistake 😅

.npmignore
ahnpnl: Also copy paste mistake

package.json
ahnpnl: This syntax is shorter than the original syntax ^x.x.x which behaves the same as the original syntax, dependabot also understands this so that is assured

ahnpnl:
Sure I will add some explanations for each script

.npmignore
ahnpnl: done

.npmignore
ahnpnl: done

ahnpnl:
added `README.md` in `scripts`.

scripts/lib/spawn-sync.js
```diff
@@ -0,0 +1,24 @@
+const processSpawnSync = require('child_process').spawnSync
```
thymikee, line 1: I'd go with `execa` for less stuff to maintain

ahnpnl: oh I never use that before, will check it out. Thanks !

e2e/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+
+## Running tests
+
+You run the tests on example apps with `yarn test:e2e` (or `npm run test:e2e)`. In `scripts/lib/paths.js`,
+`projectsToRun` must be modified if a new project is added to `e2e`.
+
+What will happen in order is as following:
+
+1. A bundle (we'll call it `[bundle]`) will be created for `jest-preset-angular` using `npm pack` (`yarn pack` is buggy).
```
thymikee, line 17: I'd suggest running from the source, to simplify setup. But running `bundle` has its own benefits, so I won't oppose too much

e2e/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Directory structure
+
+- `test-app-v10`: example app using Angular v10 with basic setup with `jest` and `jest-preset-angular`.
+- `test-app-v9`: example app using Angular v9 with basic setup with `jest` and `jest-preset-angular`.
+
+
+## Running tests
+
+You run the tests on example apps with `yarn test:e2e` (or `npm run test:e2e)`. In `scripts/lib/paths.js`,
+`projectsToRun` must be modified if a new project is added to `e2e`.
```
thymikee, line 13: could be determined automatically using `glob` package on `e2e/` directory. However, I'd much prefer using Jest's "projects" to facilitate e2e testing, which would get rid of custom sync tests running script

scripts/e2e.js
```diff
@@ -0,0 +1,64 @@
...
+
+  // then we install it in the repo
+  logger.log('ensuring all depedencies of target project are installed')
+  npm.spawnSync(['install', '--no-package-lock', '--no-shrinkwrap', '--no-save'], true,{ cwd: monorepoRealPath })
+  logger.log('installing bundled version of jest-preset-angular')
+  npm.spawnSync(['install', '--no-package-lock', '--no-shrinkwrap', '--no-save', bundle], true, { cwd: monorepoRealPath })
+
+  // then we can run the tests
+  const useYarn = existsSync(join(monorepoRealPath, 'yarn.lock'))
+  const cmdLine = projectPkg.scripts && projectPkg.scripts.test ? [useYarn ? 'yarn' : 'npm', 'test'] : ['jest']
```
thymikee, line 37: we run "npm" install to then run "yarn test"? would be nicer to pick one package manager (I suggest yarn) and stick to it. I'm fine with using "npm pack" for packing

ahnpnl: ya the part of installing other dependencies can stick with `yarn`. However, the part of installing the `.tgz` needs to stick with `npm` because `yarn` doesn't have the option to say "no change in package.json when installing this dependency"

e2e/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+    The `prepare` script will be run, so `clean` and `build` (that is why e2e tests are launched before the others, since it's building as part of the process)
+ 
+2. `npm ci` and `npm install --save-dev [bundle]` will be run in each project directory in `e2e` directory.
+3. each test suite in each example project are run according to `jest` config of that project.
+
+
+## Why
+
+E2E tests being in a temporary folder of the operating system ensure none of the parents would be containing `node_modules`.
```
thymikee, line 27: I don't see any reference to creating a temporary dir in the code. I can only see `chdir` to `e2e/<project>` and running commands there. In such case, I'd highly suggest using Jest's "projects" feature:
https://jestjs.io/docs/en/configuration#projects-arraystring--projectconfig and use Yarn workspaces to minimize the network payload when downloading all dependencies of the project and e2e tests. This may be challenging to set up, but faster builds, less maintenance and less network payloads are really nice to have

thymikee: If there's no Yarn workspace in the project now, I'd consider adding it later, after this PR is merged, because now with e2e it makes sense to use it

ahnpnl: ok let's do it in a separate PR to setup yarn workspace + jest "project"

e2e/README.md
ahnpnl: will do in a separate PR to setup yarn workspace and use jest "projects"

ahnpnl:
made a push for:

- Use `execa` when executing commands

- Install dependencies and run tests for projects with `yarn`, only installing bundle with `npm`

The rest of the comments are for another PR which is related to yarn workspace and jest `projects`.

e2e/README.md
wtho: The issue is that we have different behavior of the preset for a Angular v9 and v10 App, so we want to test both. If we use yarn workspaces and only install the dependencies once, we should not be able to install different Angular versions. Correct me if I'm wrong.

wtho: Ok actually I think this will work with yarn workspaces, but we will see in the follow-up PR.

scripts/lib/spawn-sync.js
ahnpnl: would you please check again @thymikee ?

PR #446, opened 2020-08-19
build(devs-infra): configure eslint

ahnpnl:
## Summary

- Configure eslint based on recommended **eslint/typescript-eslint** config and disable a few rules which are not necessary.

- Upgrade typescript to 3.9.7 which requires our test snapshots to be updated.

- Update `README.md` because the restructure PR didn't update it.

## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information

**N.A.**

### DIFF
--- a/src/InlineFilesTransformer.ts
+++ b/src/InlineFilesTransformer.ts
@@ -30,7 +30,7 @@ import type {
   Transformer,
   Visitor,
   PropertyAssignment,
-  Identifier
+  LiteralLikeNode,
 } from 'typescript';
 import { getCreateStringLiteral, ConfigSet } from './TransformUtils';
 
@@ -69,7 +69,7 @@ export const version = 1;
  * The factory of hoisting transformer factory
  * @internal
  */
-export function factory(cs: ConfigSet) {
+export function factory(cs: ConfigSet): (ctx: TransformationContext) => Transformer<SourceFile> {
   /**
    * Our compiler (typescript, or a module with typescript-like interface)
    */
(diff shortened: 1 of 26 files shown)


src/InlineFilesTransformer.ts
```diff
@@ -81,14 +81,8 @@ export function factory(cs: ConfigSet) {
...
-  function isPropertyAssignmentToTransform(
-    node: Node
-  ): node is PropertyAssignment {
-    return (
-      ts.isPropertyAssignment(node) &&
-      ts.isIdentifier((node as PropertyAssignment).name) &&
-      TRANSFORM_PROPS.includes(((node as PropertyAssignment).name as Identifier).text)
-    );
+  function isPropertyAssignmentToTransform(node: Node): node is PropertyAssignment {
+    return ts.isPropertyAssignment(node) && ts.isIdentifier(node.name) && TRANSFORM_PROPS.includes(node.name.text);
```
wtho, line 85: Interesting, I was getting errors without the casts

wtho (approved):
Nice! Enhances code style consistency!

src/InlineFilesTransformer.ts
ahnpnl: weird, I tested with both windows and macos, `yarn build` is fine for me. Maybe you need to remove `node_modules` and reinstall 😃

wtho: Maybe it was still on an older TS version with less inference when I originally implemented this.

PR #447, opened 2020-08-20
build(devs-infra): restructure unit test config

ahnpnl:
## Summary

- Move `tsconfig.spec.json` to root level

- Create `jest.config.js` and remove the configuration from `package.json`.

## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information
**N.A.**


jest.config.js
```diff
@@ -0,0 +1,11 @@
+/** @typedef {import('ts-jest')} */
```
ahnpnl, line 1: See https://kulshekhar.github.io/ts-jest/user/config/#ide-ts-jest-config-suggestion

jest.config.js
```diff
@@ -0,0 +1,11 @@
...
+/** @type {import('@jest/types').Config.InitialOptions} */
+module.exports = {
+  preset: 'ts-jest',
```
ahnpnl, line 4: once we have unit tests for `js` files, the preset value should be `ts-jest/presets/js-with-ts` and also add `allowJs: true` to `tsconfig.spec.json`

tsconfig.spec.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "types": ["jest", "node"]
+  },
+  "include": [] // test speed boost trick
```
wtho, line 6: Do you want to commit this test to master?

ahnpnl: oh this one means “boost test speed” 😅

ahnpnl: updated comment to be less confused

jest.config.js
ahnpnl: Or maybe we should make all files as `ts`, including serializers. My idea is: group all serializers in one folder, another folder for transformers and these folders will have an `index.ts`.

In the root index, we can link to those `index` to export them to public.

If we do so, we should do in the next major release.

ahnpnl: I think we should do one minor release which includes zone.js fix + preset config change. Then after that we can proceed with the next major

README.md
```diff
@@ -326,10 +326,12 @@ Override `globals` object in Jest config:
       "ts-jest": {
         "tsConfig": "<rootDir>/tsconfig.custom.json",
         "stringifyContentPathRegex": "\\.html$",
-        "astTransformers": [
```
ahnpnl, line 329: previous PR I didn't see this, now it is corrected

tsconfig.json
```diff
@@ -7,12 +7,12 @@
     "allowJs": true,
     "allowSyntheticDefaultImports": true,
     "noFallthroughCasesInSwitch": true,
-    "noImplicitAny": true,
```
ahnpnl, line 10: `noImplicitAny: true` is included in `strict: true`

PR #448, opened 2020-08-20
feat: add support for zone.js 0.11.1

JiaLiPassion:
zone.js 0.11.1 support jest, so all logic of `zone-patch` are no longer needed,
also zone.js 0.11.1 change the `dist` bundles to `Angular Package Format`, so the
current `require` absolute path will not work any longer.

So this PR make the following changes.
1. Add logic in `setup-jest.ts`, to load `zone.js` and `zone.js/testing` moudles,
if the jest patch is already available which means zone.js is 0.11.1+, we don't need to
import zone.js module and patch ourselves. If the jest patch is not available, we
still need to use the current logic as fallback.
2. Add e2e test for Angular 10 with zone.js 0.11.1


JiaLiPassion:
@ahnpnl , please review, thank you!

e2e/test-app-v10-zone-v11/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "test-app-v10",
```
ahnpnl, line 2: NIT, this should be `test-app-v10-zone-v11`

JiaLiPassion:
The CI is failing, but it works in my local, will check what is wrong.

src/setup-jest.ts
```diff
@@ -2,12 +2,27 @@
...
+  // If the user are using zone.js 0.11.1+
+  // all jest support logic are implemented inside zone.js
+  // we only need to load zone.js and zone.js/testing module
+  require('zone.js');
+  require('zone.js/testing');
+  if (!!jest['__zone_patch__']) {
+    builtInJestSupportInZoneJs = true;
+  }
+} catch (err) {}
+if (!builtInJestSupportInZoneJs) {
```
ahnpnl, line 16: e2e for `test-app-v10-zone-v11` failed due to this `builtInJestSupportInZoneJs` has `false` value

JiaLiPassion: Yeah, it seems the `jest` global instance are different in `zone-testing` and `setup-jest`. I will do a little more investigation.

src/setup-jest.ts
```diff
@@ -2,12 +2,27 @@
...
+try {
+  // If the user are using zone.js 0.11.1+
+  // all jest support logic are implemented inside zone.js
+  // we only need to load zone.js and zone.js/testing module
+  require('zone.js');
+  require('zone.js/testing');
+  if (!!jest['__zone_patch__']) {
+    builtInJestSupportInZoneJs = true;
+  }
+} catch (err) {}
```
splincode, line 15: maybe you should show error (for debug)?

```ts
try {
 //..
} catch (err) {
  console.log(err.message);
}
```

JiaLiPassion:
@ahnpnl , I modified the logic, we don't need to check the `__zone_patch__` flag any more, since zone.js `0.11.1` change the bundle folder structure, we just need to load it and check whether there is error. Please review, thanks!

e2e/test-app-v10-zone-v11/package.json
JiaLiPassion: Got it, updated.

ahnpnl:
thanks a lot @JiaLiPassion !!, the changes LGTM.

@wtho @thymikee would you please also take a look ?

JiaLiPassion:
@ahnpnl , thanks, still need some final tuning, the bundle name is not correct, will update in a minute.

JiaLiPassion:
@ahnpnl, I have updated the commit, now it is ok for review, thanks.

Xavier-explore-dev (approved):

src/setup-jest.ts
```diff
@@ -2,12 +2,22 @@
...
+try {
+  // If the user are using zone.js 0.11.1+
+  // all jest support logic are implemented inside zone.js
+  // we only need to load zone-testing.umd.js module
+  require('zone.js/bundles/zone-testing-bundle.umd.js');
+} catch (err) {
+  // Fallback logic to load zone and zone-patch
+  // when the user still use zone.js 0.10.x
+  // TODO: @JiaLiPassion, remove the fallback when Angular 10 LTS finished.
+  require('zone.js/dist/zone');
```
ahnpnl, line 14: I'm not sure if we can remove this fallback later because the preset supposes to support various Angular versions. If removing fallback means we don't have backward compatibility anymore for Angular < 9 ?

JiaLiPassion: Yeah, totally depends on the policy of the library, of course we can support any version of Angular with the latest version of the jest-preset-angular, or we can support different version of Angular with different versions of the jest-preset-angular. I will remove this comment.

PR #473, opened 2020-09-12
feat: replace type property with weakmap

wtho:
I recently learned about weakmaps and why using them is a much better approach to store metadata about an object than to store it as a property on that object itself.

I realized I once introduced the wrong approach in jest-preset-angular and now want to fix this.

Also I removed most `any` in `reflect-metadata` and added more typings.

### DIFF
--- a/src/utils/reflect-metadata.ts
+++ b/src/utils/reflect-metadata.ts
@@ -11,11 +11,11 @@ declare namespace Reflect {
 const METADATA_KEY_PARAMTYPES = 'design:paramtypes';
 
 // weird workaround to avoid 'ReferenceError: globalThis is not defined' in node version < 11
-// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment
-(global as any).globalThis = (global as any).globalThis || undefined;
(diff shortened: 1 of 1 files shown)


ahnpnl (approved):
I like it !

src/utils/reflect-metadata.ts
```diff
@@ -1,31 +1,41 @@
...
 // weird workaround to avoid 'ReferenceError: globalThis is not defined' in node version < 11
-// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment
-(global as any).globalThis = (global as any).globalThis || undefined;
-
-const _global = globalThis || global; // globalThis available since node v12/TS v3.4
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const reflect: any = _global['Reflect']; // reflect type in global has not these methods
+(global as any).globalThis = global.globalThis || undefined;
+
+const globalRef = globalThis || global; // globalThis available since node v12/TS v3.4
```
ahnpnl, line 17: maybe we can use `??` ?

wtho: Not in node v10

PR #524, opened 2020-10-01
refactor(presets): use `tsconfig` instead of `tsConfig` for `ts-jest`

ahnpnl:
## Summary

`ts-jest` will deprecate `tsConfig` in favor of `tsconfig`. This PR is to:

- Change all usages of `tsConfig` to `tsconfig`.

- Update `README` to instruct users to use `tsconfig` option.

## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information
**N.A.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ By Angular CLI defaults you'll have a `src/test.ts` file which will be picked up
 module.exports = {
   globals: {
     'ts-jest': {
-      tsConfig: '<rootDir>/tsconfig.spec.json',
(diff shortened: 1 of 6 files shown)


wtho (changes requested):
CHANGELOG entry is missing, otherwise LGTM

ahnpnl:
CHANGELOG is generated automatically via conventional changelog. We don't need to do it manually anymore.

wtho:
Yeah, I thought about it after I send the comment :+1:

wtho (approved):

PR #525, opened 2020-10-02
refactor(config): use `readConfiguration` from `@angular/compiler-cli`

ahnpnl:
## Summary

Introduce `NgJestConfig` which extends `ts-jest` config class `ConfigSet` and use `readConfiguration` to read tsconfig and get Angular compiler configs.

With [readConfiguration function](https://github.com/angular/angular/blob/master/packages/compiler-cli/src/perform_compile.ts#L141) we are assured that we have similar tsconfig as well as [Angular compiler options](https://angular.io/guide/angular-compiler-options) like Angular CLI does. Later this config will be required to use with Angular compiler.

Next step will be: to check which options will be overwritten in https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/angular_compiler_plugin.ts#L161

## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information
**N.A.**

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,11 +1,13 @@
 /** @typedef {import('ts-jest')} */
 /** @type {import('@jest/types').Config.InitialOptions} */
 module.exports = {
-  preset: 'ts-jest',
   globals: {
     'ts-jest': {
       tsconfig: 'tsconfig.spec.json',
     },
+  },
+  transform: {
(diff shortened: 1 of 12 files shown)


ahnpnl:
hmm I should create a branch based on master instead. Will close this and open another one with base on master

PR #526, opened 2020-10-02
refactor: introduce `AngularJestTransformer`

ahnpnl:
## Summary

Now our `index.ts` will contain entry for jest transformer. That means it is a breaking change for users who are doing
```
import 'jest-preset-angular'
```
in their `setup-jest.ts` (or equivalent).

However, the breaking change is easy to migrate. The new transformer will stay inactive for now.

- [x] Yes
- [ ] No

Users who are using `import 'jest-preset-angular'` should change to `import 'jest-preset-angular/config/setup-jest'`

## Other information
**N.A.**


ahnpnl:
@wtho I think we should cut release **9.0.0** soon so users have less things to migrate for all the breaking changes. Currently we have 3 breaking changes for **9.0.0**

- New entry for `setup-jest`

- New way of configuring serializers

- New way of configuring transformers

wtho:
Ok!
Just for clarification, is the change of `index.ts` really necessary? What exactly do we gain? What do users gain? Is it worth the breaking change?

ahnpnl:
It is necessary if we consider the basic structure of a jest transformer. A jest transformer starts with `index.ts` (more precise is `index.js`). This file will export a transformer object (or es6 class). It is considered as an entry of a jest transformer.

If we don’t follow that basic structure, we can follow the way `ts-jest` does, which is:

- transformer class is in a different file

- `index.ts` imports the transformer class then export so that `jest` can use it, see https://github.com/kulshekhar/ts-jest/blob/master/src/index.ts#L43

Besides, if we let user import `setup-jest.ts` from `index.ts` and also `index.ts` contains the transformer class, I think it might cause side effects or unknown behaviors.

IMO, it’s safer to follow the standard structure.

wtho:
Ok! LGTM!

wtho (approved):

PR #528, opened 2020-10-04
refactor(config): use `readConfiguration` from `@angular/compiler-cli`

ahnpnl:
## Summary

Introduce `NgJestConfig` which extends `ts-jest` config class `ConfigSet`. In this class, we override `readTsConfig` to use `readConfiguration` from `@angular-compiler/cli` to retrieve tsconfig options as well as Angular compiler options.

For now, this class isn't used anywhere. The plan is use it after #526 is merged.

## Test plan

Added unit tests, green CI

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information
**N.A.**

PR #535, opened 2020-10-06
refactor: use `NgJestConfig` in `AngularJestTransformer`

ahnpnl:
## Summary

Continue the work in #528

Override 2 methods of `TsJestTransformer` from `ts-jest`:
- `configsFor`: initialize our config class `NgJestConfig`.
- `getCacheKey`: use `@jest/create-cache-key-function` package and create cache key from tsconfig as well as Angular compiler options

This will ensure that later `AngularJestTransformer` can use Angular compiler with correct compiler config to compile codes.

## Test plan

Added unit tests for `AngularJestTransformer` method `configsFor`, green CI

- [ ] Yes
- [x] No

## Other information
**N.A.**


src/index.ts
```diff
@@ -1,8 +1,53 @@
...
 import type { Config } from '@jest/types';
+import createCacheKey from '@jest/create-cache-key-function';
 import { TsJestTransformer } from 'ts-jest/dist/ts-jest-transformer';
 
+import { NgJestConfig } from './config/ng-jest-config';
+import { stringify } from './utils/json';
+import { JsonableValue } from './utils/jsonable-value';
+
+interface ConfigSetItem {
+  configSet: NgJestConfig;
```
wtho, line 11: Does a `ConfigSetItem` contain a `configSet`? Isn't it more like a `ConfigSetCollection` or some other form of container/wrapper?

src/index.ts
```diff
@@ -1,8 +1,53 @@
...
+   * @internal
+   */
+  configsFor(jestConfig: Config.ProjectConfig): NgJestConfig {
+    let csi: ConfigSetItem | undefined = AngularJestTransformer.configSetItems.find(
+      (cs) => cs.jestConfig.value === jestConfig,
+    );
+    if (csi) return csi.configSet;
+    // try to look-it up by stringified version
+    const serialized = stringify(jestConfig);
+    csi = AngularJestTransformer.configSetItems.find((cs) => cs.jestConfig.serialized === serialized);
```
wtho, line 33: We can avoid the upper `let` if we introduce a second variable for `csi`, like `const deserializedCsi = AngularJestTransformer.configSetItems...`

src/index.ts
```diff
@@ -1,8 +1,53 @@
...
+    if (csi) {
+      // update the object so that we can find it later
+      // this happens because jest first calls getCacheKey with stringified version of
+      // the config, and then it calls the transformer with the proper object
```
wtho, line 37: I think I need more information about this edge case :smile:

src/utils/jsonable-value.ts
```diff
@@ -0,0 +1,35 @@
+import { stringify } from './json';
+
+/**
+ * Copy from `ts-jest`
+ * @internal
+ */
+export class JsonableValue<V = Record<string, unknown>> {
```
wtho, line 7: Can we rename it to `JsonObject`?

This use case for this class is for caching, isn't it?

src/index.ts
ahnpnl: It is a like a container/wrapper. The purpose of this is whenever our jest transformer discovers a processing jest config which is already inside `configSetITems`, we simply return the cached instance `NgJestConfig` without initializing a new instance.

src/index.ts
ahnpnl: This one handles the case that when cannot find any jest config by memory reference in `configSetItems`, we fallback to find by serialized string. If match, we simply update memory reference. This case hardly happens, just for safety.

src/utils/jsonable-value.ts
ahnpnl: yes, we can rename it. It is mainly for caching.

src/index.ts
ahnpnl: Do you mean we don't need the `let csi` or do you mean 
```
const csi = AngularJestTransformer.configSetItems.find(
      (cs) => cs.jestConfig.value === jestConfig,
    );
if (csi) return csi.configSet;
// try to look-it up by stringified version
const serialized = stringify(jestConfig);
const deseralizedCsi = AngularJestTransformer.configSetItems...
```

wtho: The second one :+1:

src/index.ts
wtho: Ok, so what about `ConfigSetCache` or `ConfigSetStore`?

ahnpnl: LGTM, I will change that

ahnpnl:
I added couple of `TODO` which `ts-jest` can expose for us so we don't have to maintain those.

src/index.ts
```diff
@@ -11,6 +32,55 @@ class AngularJestTransformer extends TsJestTransformer implements Transformer {
...
+  ): string {
+    this.createOrResolveTransformerCfg(transformOptions.config);
+
+    return createCacheKey()(fileContent, filePath, this._ngJestCfgStr, {
+      config: transformOptions.config,
+      instrument: false,
+    });
+  }
+
+  private createOrResolveTransformerCfg(jestConfig: Config.ProjectConfig): void {
```
ahnpnl, line 50: @wtho I added this new method and remove `configsFor`. The purpose is because `_ngJestCfgStr` is only scoped to this class so I don't want to put it into `NgJestConfig` like originally `ts-jest` does with its cache key.

src/index.ts
wtho: Hmm, basically I do not like `CachedConfigSet.configSet` so much. It's like `Car.car`. The configSet itself should already be the reference on the set.

src/index.ts
wtho: Looks good!

src/index.ts
ahnpnl: Ok let’s use the name `ConfigSetStore`

wtho: No, actually it's good for me this way.

If you don't mind we just rename the inner `configSet: NgJestConfig` to `ngJestConfig: NgJestConfig`?

ahnpnl: hmm, maybe like this
```
interface ConfigSetItem {
    ngJestConfig: NgJestConfig;
    jestConfig: JsonObject<Config.ProjectConfig>;
    transformerConfig: JsonObject<Config.ProjectConfig & ParsedConfiguration>; // this is for jest cache key
}
```

wtho: IMO this is better

ahnpnl:
done, I think everything is good to go now

wtho (approved):
One last nit: rename of `src/__tests__/jsonable-value.spec.ts` to `src/__tests__/json-object.spec.ts`

Otherwise great work!
Sorry for all the nits and the discussion, but I think doing these improves the quality from the beginning.
Thanks for all the work in this!

ahnpnl:
> One last nit: rename of `src/__tests__/jsonable-value.spec.ts` to `src/__tests__/json-object.spec.ts`
> 
> Otherwise great work!
> Sorry for all the nits and the discussion, but I think doing these improves the quality from the beginning.
> Thanks for all the work in this!

Actually I gonna remove that test file instead because this JsonObject class is a temporary copy from `ts-jest`. It will be removed when `ts-jest` exposes it so this test file is not needed.

Issue #353, opened 2020-02-29
entryComponents with Angular 9 (Ivy)


timdeschryver:
With Angular 9 (Ivy), we no longer have to provide `entryComponents`.
This seems to fail if we're using Jest as our test runner.
The same test does pass with Jasmine/Karma.

If we run it with Jest, we're getting the following error

```
No component factory found for DynamicButtonComponent. Did you add it to @NgModule.entryComponents?

      at noComponentFactoryError (../packages/core/src/linker/component_factory_resolver.ts:17:17)
      at CodegenComponentFactoryResolver.Object.<anonymous>.CodegenComponentFactoryResolver.resolveComponentFactory (../packages/core/src/linker/component_factory_resolver.ts:72:13)
      at NgComponentOutlet.Object.<anonymous>.NgComponentOutlet.ngOnChanges (../packages/common/src/directives/ng_component_outlet.ts:105:36)
      at checkAndUpdateDirectiveInline (../packages/core/src/view/provider.ts:207:15)
      at checkAndUpdateNodeInline (../packages/core/src/view/view.ts:429:14)
      at checkAndUpdateNode (../packages/core/src/view/view.ts:389:12)
      at debugCheckAndUpdateNode (../packages/core/src/view/services.ts:430:44)
      at debugCheckDirectivesFn (../packages/core/src/view/services.ts:391:7)
      at Object.eval [as updateDirectives] (ng:/DynamicHostModule/DynamicHostComponent.ngfactory.js:14:9)
      at Object.debugUpdateDirectives [as updateDirectives] (../packages/core/src/view/services.ts:385:19)
      at checkAndUpdateView (../packages/core/src/view/view.ts:359:12)
      at callViewAction (../packages/core/src/view/view.ts:615:11)
      at execComponentViewsAction (../packages/core/src/view/view.ts:559:7)
      at checkAndUpdateView (../packages/core/src/view/view.ts:370:3)
      at callWithDebugContext (../packages/core/src/view/services.ts:629:23)
      at Object.debugCheckAndUpdateView [as checkAndUpdateView] (../packages/core/src/view/services.ts:346:10)
      at ViewRef_.Object.<anonymous>.ViewRef_.detectChanges (../packages/core/src/view/refs.ts:260:16)
      at ComponentFixture.Object.<anonymous>.ComponentFixture._tick (../../packages/core/testing/src/component_fixture.ts:107:28)
      at ../../packages/core/testing/src/component_fixture.ts:120:36
[...]

PR #543, opened 2020-10-08
Compiler

ahnpnl:
This branch managed to fix the issue with entry component #353 but it failed on other tests, some tests even hang.


wtho:
What's the status? Do you have an idea why tests tend to hang?

ahnpnl:
after `ngcc`, running tests hang due to `Converting circular structure to JSON` exception. I haven't looked into it more yet. It also can be that the compiler codes are not completed, not so sure.

ahnpnl:
I will create another draft PR. The branch on this PR is a bit messy

PR #553, opened 2020-10-17
test(e2e): improve e2e tests

ahnpnl:
## Summary

- Fix exposed paths for transformers
- Copy `jest-preset.js`, `build` manually to each test app
- Always trigger build assets before starting e2e test
- Correct `README` related to `setup-jest`

## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information
**N.A.**


ahnpnl:
@wtho now our e2e tests have exactly same setup as normal projects which install the preset.

wtho (approved):
Ok, awesome! Wanted to have a realistic test setup since such a long time! Thanks!

It might just be a bit confusing for devs who look at our test projects to not have `jest-preset-angular` in dependencies. But then again, these are test projects, not examples.

:+1:

ahnpnl:
Our test projects now look exactly how end users use the preset, so the confusion is gone now :)

PR #612, opened 2020-11-07
test(e2e): add custom typing e2e test

ahnpnl:
## Summary

Add custom typing e2e test

## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information
**N.A.**

### DIFF
--- a/e2e/test-app-v10-zone-v11/tsconfig.spec.json
+++ b/e2e/test-app-v10-zone-v11/tsconfig.spec.json
@@ -6,9 +6,13 @@
     "module": "commonjs",
     "types": [
       "jest",
+      "jquery",
       "node"
     ]
   },
+  "files": [
+    "src/global.ts"
+  ],
   "include": [
     "src/**/*.spec.ts",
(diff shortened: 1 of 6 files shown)


e2e/test-app-v10-zone-v11/tsconfig.spec.json
```diff
@@ -6,9 +6,13 @@
     "module": "commonjs",
     "types": [
       "jest",
+      "jquery",
       "node"
     ]
   },
+  "files": [
+    "src/global.ts"
```
wtho, line 14: Interesting, I didn't know "global" types could be added like this.

ahnpnl: Actually I don't like this way. I'd prefer more using `d.ts` instead, but ya it's possible indeed.

wtho: Yeah, `d.ts` and explicit `import`s are definitively better

PR #620, opened 2020-11-10
build(devs-infra): build package with target `es2015`

ahnpnl:
## Summary

Ship package with target `es2015`. This is more inline with jest, see https://jestjs.io/blog/2020/05/05/jest-26 as well as Angular

## Test plan

Green CI

- [x] Yes
- [ ] No

`jest-preset-angular` no longer ships with `es5` build but now with `es2015`

## Other information
**N.A.**


thymikee (approved):
👍

PR #621, opened 2020-11-10
refactor(preset): explicitly define `jsdom` for test environment

ahnpnl:
## Summary

Jest 27 sets default test environment to `node`, see https://github.com/facebook/jest/pull/9874. Since we need `jsdom` so we need to explicitly set it in `jest-preset.js`.

Jest 27 will still bundle `jest-environment-jsdom` in the main package and Jest 28 will separate them. However, we should just go ahead and set explicitly `jest-environment-jsdom` in our dependencies list.

More information see https://jestjs.io/blog/2020/05/05/jest-26

## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information
**N.A>**

### DIFF
--- a/jest-preset.js
+++ b/jest-preset.js
@@ -11,6 +11,7 @@ module.exports = {
       },
     },
   },
+  testEnvironment: 'jsdom',
   transform: {
     '^.+\\.(ts|js|html)$': 'ts-jest',
   },
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   },
   "dependencies": {
     "@jest/create-cache-key-function": "^26.5.0",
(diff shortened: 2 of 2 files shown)


jest-preset.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
       },
     },
   },
+  testEnvironment: 'jest-environment-jsdom-sixteen',
```
thymikee, line 14: isn't latest Jest using jsdom v16?

ahnpnl: for some reasons, my latest push isn't affected

ahnpnl: should be good now :)

thymikee (approved):

Issue #764, opened 2021-01-28
File changes not getting picked up in --watch mode (9.0.0-next only)


johncrim:
## 🐛 Bug Report

When using 9.0.0-next.6 with isolatedModules: false, in  `jest --watch` file changes are not being properly incorporated - the file state when jest is started is what is run, regardless of changes. Jest appears to rebuild the files on changes, but is not running the new code.

This is not an issue with islatedModules: true.

I've debugged this into `NgJestCompiler` and verified that the `ts.SourceFile` is not being updated with the new `fileContent`, so I will provide a PR to fix this shortly.

## To Reproduce

If I have a test like the following, which fails to compile b/c the commented out line is missing

```ts
describe('A directive can be declared in the test file', () => {

  it('and used in a test', () => {
    TestBed.configureTestingModule({
      declarations: [BlockClickBubbleDirective, TestHostComponent]
    });
    //const fixture = TestBed.createComponent(TestHostComponent);

    const directive = fixture.debugElement.children[0].children[0].injector.get(BlockClickBubbleDirective);
    expect(directive).toBeTruthy();
  })

});
```

then run `jest --watch`, I receive an error (as expected). If I then uncomment the `const fixture` line, the file is rebuilt and the test is re-run, but it uses the original (erroring) code and I receive the original error. If I exit jest and restart, the corrected code is compiled and run.

## Expected behavior

In `--watch` mode, code changes are compiled and run after the file is saved.

## Link to repo

https://github.com/johncrim/repro-ivy-jest-preset-angular/tree/bug/jest-watch-update

```
git clone git@github.com:johncrim/repro-ivy-jest-preset-angular.git
git checkout  bug/jest-watch-update

yarn
yarn test:watch inline-directive # text failure expected
# uncomment 'const fixture' line; test re-runs and still fails
# exit + restart
yarn test:watch inline-directive # test passes
```

## envinfo

```
System:
    OS: Windows

Npm packages:
    jest: 27.0.0-next.2
    jest-preset-angular: 9.0.0-next.6
    typescript: 4.0.5
```

PR #765, opened 2021-01-28
fix(watch mode): update SourceFile with fileContent

johncrim:
Previously `SourceFile.text` was the old contents of the file - the state of the file during startup. This change updates the SourceFile text so that emit() receives the changed file contents.

fixes: #764 

## Test plan

I've tested this locally by copying the `build` dir into node_modules on 2 different projects (test repo and production repo) and verified. It would be ideal to create a unit test for this, but I'm not sure I'm up to that.

- [ ] Yes
- [x] No

### DIFF
--- a/src/compiler/ng-jest-compiler.ts
+++ b/src/compiler/ng-jest-compiler.ts
@@ -4,6 +4,7 @@ import type { Logger } from 'bs-logger';
 import { updateOutput } from 'ts-jest/dist/compiler/compiler-utils';
 import type { CompilerInstance, TTypeScript, ResolvedModulesMap } from 'ts-jest/dist/types';
 import type * as ts from 'typescript';
+import { TextSpan } from 'typescript';
 
 import type { NgJestConfig } from '../config/ng-jest-config';
 import { constructorParametersDownlevelTransform } from '../transformers/downlevel-ctor';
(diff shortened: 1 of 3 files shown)


ahnpnl (commented):
unit test should updated though, a bit similar to this test https://github.com/thymikee/jest-preset-angular/blob/master/src/__tests__/ng-jest-compiler.spec.ts#L105

src/compiler/ng-jest-compiler.ts
```diff
@@ -70,20 +71,27 @@ export class NgJestCompiler implements CompilerInstance {
       esModuleInterop,
       module: moduleKind,
     };
+
     if (this._program) {
-      const allDiagnostics: ts.Diagnostic[] = [];
-      if (!this._rootNames.includes(fileName)) {
-        this._logger.debug({ fileName }, 'getCompiledOutput: update memory host, rootFiles and Program');
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      this._tsHost!.updateMemoryHost(fileName, fileContent);
```
johncrim, line 77: I'm pretty sure this update should be made regardless of whether fileName is already included in _rootNames. Lmk if that's incorrect @ahnpnl .

johncrim:
ok, I'll give a test a shot.

src/compiler/ng-jest-compiler.ts
ahnpnl: yes this should be always done regardless `_rootNames`

src/compiler/ng-jest-compiler.ts
```diff
@@ -4,6 +4,7 @@ import type { Logger } from 'bs-logger';
 import { updateOutput } from 'ts-jest/dist/compiler/compiler-utils';
 import type { CompilerInstance, TTypeScript, ResolvedModulesMap } from 'ts-jest/dist/types';
 import type * as ts from 'typescript';
+import { TextSpan } from 'typescript';
```
ahnpnl, line 7: this import should not be needed since we already had the import star above.

ahnpnl (commented):
only NIT, the rest LGTM

PR #955, opened 2021-06-10
fix: widen range for optional webpack dependency

sonallux:
## Summary

When installing `jest-preset-angular` with angular 12.0.3, two webpack versions are installed in node_modules:
- webpack `5.36.2` from `jest-preset-angular`
- webapck `5.38.1` from `@angular-devkit/build-angular`

This PR widens the range of the optional `webpack` dependency, so the package manager can deduplicate the webpack version. This will result in faster install times and smaller node_modules folder.

## Test plan

All tests are green.

- [ ] Yes
- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "jest": "^27.0.0"
   },
   "optionalDependencies": {
-    "webpack": "5.36.2"
+    "webpack": "^5.36.2"
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "^0.1102.14",
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "jest": "^27.0.0"
   },
   "optionalDependencies": {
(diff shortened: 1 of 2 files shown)


ahnpnl:
Angular 9 10 (possibly 11 too) don’t have webpack 5. I think it’s better put something like `4.x ll 5.x`

sonallux:
> Angular 9 10 (possibly 11 too) don’t have webpack 5. I think it’s better put something like `^4.0.0 || ^5.30.0` or just `>=4.0.0`

I wasn't aware that Angular 9, 10 and 11 are also still supported. I am going to change the range to `^4.0.0 || ^5.30.0`.

ahnpnl (approved):
Thanks!!

sonallux:
@ahnpnl Can you perform a new release, so everyone can benefit from this small improvement?

Issue #1222, opened 2021-12-14
[Bug]: Errors running v13 example with ESM


johncrim:
### Version

main (v11.0.1)

### Steps to reproduce

1. git clone git@github.com:thymikee/jest-preset-angular.git
2. cd examples\example-app-v13
3. yarn
4. yarn test-esm

### Expected behavior

Example tests run without error.

### Actual behavior

All tests fail. Most tests fail with this error:

```
 FAIL  src/app/demo/async-helper.spec.ts
  ● Test suite failed to run

    Must use import to load ES Module: C:\src\github\angular\jest-preset-angular\examples\example-app-v13\node_modules\@angular\core\fesm2015\testing.mjs

      at Runtime.requireModule (node_modules/jest-runtime/build/index.js:972:21)
```

A few tests also have the error: `ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.`

```
 FAIL  src/app/demo/demo.testbed.spec.ts
  ● Test suite failed to run

    Must use import to load ES Module: C:\src\github\angular\jest-preset-angular\examples\example-app-v13\node_modules\@angular\core\fesm2015\testing.mjs

      at Runtime.requireModule (node_modules/jest-runtime/build/index.js:972:21)


ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From src/app/demo/async-helper.spec.ts.

          at async Promise.all (index 1)
```

### Environment

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
  Binaries:
    Node: 16.12.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.1.0 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    jest: ^27.4.4 => 27.4.4
```

johncrim:
If I change the setup-jest.ts file to use ESM imports (and skip using `jest-preset-angular/setup-jest`), most of the tests pass:

```ts
// examples/example-app-v13/setup-jest.ts

// Removed:
// import 'jest-preset-angular/setup-jest';

import 'zone.js/fesm2015/zone-testing-bundle.min.js';
import './jest-global-mocks';

import { getTestBed } from '@angular/core/testing';
[...]

PR #1277, opened 2022-01-17
Fix ng13 esm example

johncrim:
## Summary

Fixes: #1222 

Updates the ng13 example so testing can be run using ES modules. This may not be the preferred fix, since per @ahnpnl it's not yet possible to use a setup-jest.mjs in the preset that is run automatically when ESM mode is enabled; but it does provide a working example for developers who want to use jest, ng13, and ESM now.

## Test plan

Running `yarn test-examples` succeeds. Since I don't know if this is in the CI, both of these work:

```
yarn
yarn test-examples
```

```
cd examples\example-app-v13
yarn
yarn test-esm # all tests pass
yarn test-esm-isolated # all tests pass
```

- [ ] Yes
- [x] No


johncrim:
I also updated the docs. Lmk if you'd like me to remove the doc changes - it's certainly possible that this isn't ready for prime-time yet.

.vscode/launch.json
```diff
@@ -0,0 +1,60 @@
+{
```
ahnpnl, line 1: this file should be removed :)

examples/example-app-v13/setup-jest.ts
```diff
@@ -1,2 +1,14 @@
...
+
+import 'zone.js/fesm2015/zone-testing-bundle.min.js';
 import './jest-global-mocks';
+
+import { getTestBed } from '@angular/core/testing';
+import { BrowserDynamicTestingModule, platformBrowserDynamicTesting } from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().initTestEnvironment(BrowserDynamicTestingModule, platformBrowserDynamicTesting(), {
+  teardown: {
+    destroyAfterEach: true,
```
ahnpnl, line 12: this one is default set in Angular 13, we don't need to set here

package.json
```diff
@@ -33,8 +33,8 @@
     "postinstall": "husky install",
     "lint": "eslint --ext .js,.ts .",
     "lint-fix": "eslint --fix --ext .js,.ts .",
-    "lint-prettier": "prettier '**/*.{yml,yaml,md}' --write",
-    "lint-prettier-ci": "prettier '**/*.{yml,yaml,md}' --check",
+    "lint-prettier": "prettier \"**/*.{yml,yaml,md}\" --write",
```
ahnpnl, line 36: I think these 2 should be reverted?

website/docs/guides/angular-13+.md
```diff
@@ -5,7 +5,7 @@ title: Angular >=13
```
ahnpnl, line 5: Would you please also apply the same change for https://github.com/thymikee/jest-preset-angular/blob/main/website/versioned_docs/version-11.x/guides/angular-13%2B.md

.vscode/launch.json
johncrim: ok - apologies - I didn't intend to commit it.

examples/example-app-v13/setup-jest.ts
johncrim: Are you sure about this? My understanding is that this is set in new projects (in the karma setup file), and is recommended, but it's not a default. I'll step through to figure that out - and if you're right, I'll also remove from docs.

ahnpnl: Ye it's a part of Angular 13 change, you can see it here https://github.com/ngrx/platform/issues/3243#issuecomment-973038378

package.json
johncrim: These changes were necessary to run prettier on windows (with yarn 3). On the command line I can:

```
yarn prettier '**/*.{yml,yaml,md}' --write
```

But for some reason the commit hook always fails on this line with the previous code, but works with the change. I figure others could have the same problem, which is why I included this change. If you'd like it reverted, I'm happy to, but I will have to commit with `--no-verify`.

ahnpnl: hmm this is odd. If so, I think it's good to have.

examples/example-app-v13/setup-jest.ts
johncrim: You're right; I'll remove.

johncrim:
@ahnpnl - I've addressed your feedback (thanks!). Lmk if you see anything else.

ahnpnl (approved):
Thanks!

PR #1309, opened 2022-02-04
perf(transformer): cache useNativeEsbuild result

sorin-davidoi:
## Summary

Since the result will be the same for all transformer instances we can cache the result, thus avoiding resolving and spawning redundant processes.

For a test harness with ~3800 tests grouped in ~110 projects this translates to a reduction in running time of ~20s. Attaching a screenshot of the profile that shows how instantiation takes ~20s.

image

## Test plan

The existing tests are passing (`npm test`). Our existing test harness runs without any issues with this patch.

- [ ] Yes
- [x] No

### DIFF
--- a/src/ng-jest-transformer.ts
+++ b/src/ng-jest-transformer.ts
@@ -11,6 +11,11 @@ import type { ProjectConfigTsJest, TransformOptionsTsJest } from 'ts-jest/dist/t
 import { NgJestCompiler } from './compiler/ng-jest-compiler';
 import { NgJestConfig } from './config/ng-jest-config';
 
+// Cache the result between multiple transformer instances
+// to avoid spawning multiple processes (which can have a major
+// performance impact when used with multiple projects).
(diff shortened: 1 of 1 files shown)


src/ng-jest-transformer.ts
```diff
@@ -11,6 +11,11 @@ import type { ProjectConfigTsJest, TransformOptionsTsJest } from 'ts-jest/dist/t
 import { NgJestCompiler } from './compiler/ng-jest-compiler';
 import { NgJestConfig } from './config/ng-jest-config';
 
+// Cache the result between multiple transformer instances
+// to avoid spawning multiple processes (which can have a major
+// performance impact when used with multiple projects).
+let useNativeEsbuild: boolean | null = null;
```
ahnpnl, line 17: Can we use `boolean` only?

ahnpnl (approved):
thanks!

PR #1310, opened 2022-02-04
perf(transformer): only construct configSet when necessary

sorin-davidoi:
## Summary

The constructor of `ConfigSet` is quite expensive (path concatenations, merging configurations, reading from disk...). It should only be triggered when needed.

## Test plan

`npm test` is passing. Our test harness runs without any issues with this patch.

- [ ] Yes
- [x] No


ahnpnl:
This change is not correct. We rely on `ts-jest` to process ts files which includes TS LanguageService to be able to use Angular AST transformers so we always need `ConfigSet`. 

We only use esbuild to process `.mjs` files from `node_modules`.

sorin-davidoi:
Ah, wasn't aware of that. Closing.

PR #1313, opened 2022-02-08
perf(transformer): cache config set between runs, fixes #1311

### LINKED ISSUE #1311: [Bug]: Upgrading jest-preset-angular to 11.* makes it unusable
### Version

11.1.0

### Steps to reproduce

We have quite large app, after upgrading to 11.1.0 our tests increased by 2x in time only with jest-preset-angular upgrade.
I suspect that using lodash-es, lodash-decorators-esm, date-fns, @uppy makes it longer with the latest version. I already tried several approaches:
- align our configuration to v11 changes
- using v11 with angular 12
- running on clean configurations taken from exaples/ dir

The result is almost always the same, run time increased 2x on ci environments but even worse on my local machine. With v10 warmup took more or less 15-20 seconds, with v11 warmup takes more than 140 - 200 seconds. 

At this point using library v11 is complete blocker and blocks us from upgrading to angular 13.

### Expected behavior

Performance should stay as it was in v10

### Actual behavior

v11 increases built time more than 2x.

### Additional context

Our tests with
Test Suites: 1 skipped, 349 passed, 349 of 350 total
Tests:       3 skipped, 1247 passed, 1250 total

v10.1 with clean cache:
Time:        181.311 s

With cache:
Time:        84.668 s

v11.1.0
Time:        421.699 s

With cache: 
Time:        132.581 s

```jest.config.js
require('jest-preset-angular/ngcc-jest-processor');
const { pathsToModuleNameMapper } = require('ts-jest');
const { compilerOptions } = require('./tsconfig.json');

process.env.TZ = 'Europe/Warsaw';

module.exports = {
    preset: 'jest-preset-angular',
    testMatch: ['**/*.spec.ts'],
    maxWorkers: '50%',
    setupFilesAfterEnv: ['<rootDir>/jest.setup.ts'],
    moduleNameMapper: {
        '^lodash-es/(.*)$': 'lodash/$1',
        ...pathsToModuleNameMapper(compilerOptions.paths, { prefix: '<rootDir>/src/' }),
    },
    testResultsProcessor: './node_modules/jest-junit-reporter',
    collectCoverageFrom: ['src/**/*.ts'],
    coverageReporters: ['json', 'html'],
    transformIgnorePatterns: ['<rootDir>/node_modules/(?!lodash-es|lodash-decorators-esm|date-fns|@uppy/utils)'],
    globals: {
        'ts-jest': {
            stringifyContentPathRegex: '\\.(html|svg)$',
            isolatedModules: true,
            tsconfig: './tsconfig.spec.json',
        },
    },
};

```

### Environment

```shell
System:
    MacOS 12.1
Binaries:
    Node 14.15.3
    NPM 6.14.9
Packages:
    Jest: 27.4.7
    @ngneat/spectator 7.2.0
    Angular 12.2.14 && Angular 13
```

yadue:
## Summary

Performance fix for 11. Caches configSet between runs

## Test plan
The existing tests are passing (npm test). Our existing test harness runs without any issues with this patch.

- [ ] Yes
- [x] No

## Other information
Fixes: https://github.com/thymikee/jest-preset-angular/issues/1311


wtho (dismissed):

ahnpnl:
One thing I'm not certain about is: the caching of `ConfigSet` is supposed to be done by `ts-jest` https://github.com/kulshekhar/ts-jest/blob/main/src/ts-jest-transformer.ts#L68 but it seems like it doesn't. The fix looks ok, but it might cause side effects though.

I'd suggest to investigate more why `ts-jest` doesn't cache it.

yadue:
@ahnpnl caching is being done internally in ts-jest-transfomer, but here we dont use it but forcing generation each time. This method is also private and all fields.

ahnpnl:
Actually not, we override create config set because `ts-jest` will call it when cache isn’t found.

yadue:
@ahnpnl yeah it works fine when its already cached, but when its not we're regenerating configSet over and over. Ts-jest does this in a proper way, caches configSet it and reuses. Here, if there is no cache config set is being generated milion times. Each configSet takes 150ms on my machine and its being run for all used files multiplied by number of workers.

src/ng-jest-transformer.ts
```diff
@@ -50,7 +51,7 @@ export class NgJestTransformer extends TsJestTransformer {
     filePath: Config.Path,
     transformOptions: TransformOptionsTsJest,
   ): TransformedSource | string {
-    const configSet = this._createConfigSet(transformOptions.config);
+    this.#configSet = this.#configSet ? this.#configSet : this._createConfigSet(transformOptions.config);
```
ahnpnl, line 54: Here if you use `this._configsFor` would allow `ts-jest` to do caching on our `NgJestConfig`. You can use `@ts-expect-error` with comment saying that `ts-jest` doesn’t expose the method as public typings.

yadue: unfortunately it doesn't work properly

```
  TypeError: Cannot read property 'get' of undefined

       9 | export class NgJestCompiler extends TsCompiler {
      10 |   constructor(readonly configSet: ConfigSet, readonly jestCacheFS: Map<string, string>) {
    > 11 |     super(configSet, jestCacheFS);

```

anyway you know whats the problem and how to solve it, feel free to reject this PR and prepare correct solution, we already spend more time than you would spend doing that alone.. I need to get back to my primary work and looking forward for a fix. thanks in advance

yadue:
@ahnpnl @wtho I prepared one last change based on original code. It should cache it like original function. I coudn't use this._configsFor since it makes additional issues. I'm leaving the rest to you guys.

src/ng-jest-transformer.ts
sorin-davidoi: This is because the mocked configuration is missing the cache, opened https://github.com/thymikee/jest-preset-angular/pull/1318.

ahnpnl: the #1318 contains exactly what I was talking about :) thanks @sorin-davidoi

Issue #1466, opened 2022-05-11
[Feature]: Pass destroyAfterEach in the config


chabb:
### 🚀 Feature Proposal

Is it possible to pass  `destroyAfterEach` in the configuration file ? 
It does not seem to be the case without explicitly doing: 

```
getTestBed().initTestEnvironment(BrowserDynamicTestingModule, platformBrowserDynamicTesting(), {
  teardown: {
    destroyAfterEach: true, // Angular recommends this, but it may break existing tests
  },
});
```

### Motivation

Our tests fail on angular12,  if this flag is set to false; and it's false by default on angular 12

PR #1475, opened 2022-05-16
feat: allow passing the whole ModuleTeardownOptions

chabb:
## Summary

In  #1466 , I've forgotten to mention that there was a whole configuration object for the teardown options. 

- [ ] Yes
- [x] No

People who passed the `destroyAfterEach` flag will need to pass the whole options object

### DIFF
--- a/setup-jest.mjs
+++ b/setup-jest.mjs
@@ -2,12 +2,10 @@ import 'zone.js/fesm2015/zone-testing-bundle.min.js';
 import { getTestBed } from '@angular/core/testing';
 import { BrowserDynamicTestingModule, platformBrowserDynamicTesting } from '@angular/platform-browser-dynamic/testing';
 
-const configuredDestroyAfterEach = globalThis.ngJest?.destroyAfterEach;
-if (configuredDestroyAfterEach !== undefined) {
+const teardown = globalThis.ngJest?.teardown;
+if (teardown !== undefined) {
   getTestBed().initTestEnvironment(BrowserDynamicTestingModule, platformBrowserDynamicTesting(), {
-    teardown: {
-      destroyAfterEach: configuredDestroyAfterEach,
-    },
+    teardown,
   });
 } else {
   getTestBed().initTestEnvironment(BrowserDynamicTestingModule, platformBrowserDynamicTesting());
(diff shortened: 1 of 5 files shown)


ahnpnl:
Can we make it to not be a breaking change?

chabb:
I've added a commit to let the user pass `destroyAfterEach`

src/global.d.ts
```diff
@@ -1,10 +1,12 @@
 /* eslint-disable */
 
+import {ModuleTeardownOptions} from "@angular/core/testing";
```
ahnpnl, line 3:
```suggestion
import type { ModuleTeardownOptions } from "@angular/core/testing";
```

src/global.d.ts
```diff
@@ -1,10 +1,12 @@
...
+
 declare global {
   var ngJest: {
     skipNgcc?: boolean;
     tsconfig?: string;
-    destroyAfterEach?: boolean;
```
ahnpnl, line 7: The property `destroyAfterEach` should still remain until we drop it

setup-jest.mjs
```diff
@@ -2,12 +2,18 @@ import 'zone.js/fesm2015/zone-testing-bundle.min.js';
 import { getTestBed } from '@angular/core/testing';
 import { BrowserDynamicTestingModule, platformBrowserDynamicTesting } from '@angular/platform-browser-dynamic/testing';
 
+let teardown = globalThis.ngJest?.teardown;
 const configuredDestroyAfterEach = globalThis.ngJest?.destroyAfterEach;
-if (configuredDestroyAfterEach !== undefined) {
+
+if (configuredDestroyAfterEach) {
```
ahnpnl, line 8: I think we can add a `console.warn` to tell users to migrate to `teardown` option. Also inform users that `destroyAfterEach` is deprecated.

ahnpnl (changes requested):

src/config/setup-jest.spec.ts
```diff
@@ -48,12 +48,33 @@ describe('setup-jest', () => {
   beforeEach(() => {
     delete globalThis.ngJest;
     jest.clearAllMocks();
+    jest.resetModules();
   });
 
   test('should initialize test environment with getTestBed() and initTestEnvironment() for CJS setup-jest', async () => {
+    globalThis.ngJest = {
```
ahnpnl, line 55: The test case title should reflect about we are using `teardown` option.

setup-jest.mjs
chabb: Done. My english is not perfect, so the wording might be a bit akward. Feel free to correct if needed

ahnpnl (approved):
thanks!

PR #1655, opened 2022-06-24
Added SVG to transform config explanation

mtpultz:
Transform property doesn't explicitly mention SVG, but it is a file transform provided in the RegExp.

## Summary

Update to the documentation to include SVG in the file transform config description, which is included in the preceding RegExp of the transform property in the example object literal.

## Test plan

None.  Documentation update.

- [ ] Yes
- [X] No


ahnpnl:
Would you please also modify for these locations:
- https://github.com/thymikee/jest-preset-angular/tree/main/website/docs
- all possible doc versions under https://github.com/thymikee/jest-preset-angular/tree/main/website/versioned_docs If any versions mentioned about supporting `SVG` in config, the explanations need to be updated.

PR #1659, opened 2022-06-25
Allows default and named export for global-setup

rafagsiqueira:
## Summary
If someone wants to run another function before calling the global-setup, they can use a named export or a default export for importing the original global-setup function and calling it manually.
## Test plan
Jest already supports default exports for globalSetup functions, so no impact is expected.
- [ ] Yes
- [X] No

### DIFF
--- a/src/config/global-setup.ts
+++ b/src/config/global-setup.ts
@@ -1,6 +1,6 @@
 import { runNgccJestProcessor } from '../utils/ngcc-jest-processor';
 
-const GlobalSetup = () => {
+const globalSetup = async () => {
   const ngJestConfig = globalThis.ngJest;
   const tsconfig = ngJestConfig?.tsconfig;
   if (!ngJestConfig?.skipNgcc) {
@@ -8,4 +8,4 @@ const GlobalSetup = () => {
   }
 }
 
-export default GlobalSetup;
+export default globalSetup;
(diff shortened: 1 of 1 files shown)


src/config/global-setup.ts
```diff
@@ -1,9 +1,11 @@
 import { runNgccJestProcessor } from '../utils/ngcc-jest-processor';
 
-export = async () => {
+const GlobalSetup = () => {
```
ahnpnl, line 3:
```suggestion
const globalSetup = () => {
```

ahnpnl (changes requested):

ahnpnl (approved):
Thanks!

ahnpnl:
Note that, if you do in your global setup file
```
const globalSetup = require('jest-preset-angular/global-setup');
```
it will try to import this file https://github.com/thymikee/jest-preset-angular/blob/main/global-setup.js

This won't work because this file is CJS file therefore Jest will throw error unless you run test in ESM mode. 

If you run test in CJS mode, you have to do
```
const globalSetup = require('jest-preset-angular/build/config/global-setup');
```

rafagsiqueira:
I can confirm this works for a typescript module. Thank you!
```
import { globalSetup } from 'jest-preset-angular/build/config/global-setup';

async function setup() {
    await globalSetup();
    (...)
}

export default setup;
```

PR #1841, opened 2022-11-23
Change jest dependency policy

vadjs:
## Summary

This change will make possible to upgrade `jest` without upgrading `jest-preset-angular`.

Currently we have to release newer version of `jest-preset-angular` every time when new version of jest is released. The current situation is a good example: there is no possibility to upgrade jest but no new version of `jest-preset-angular` available ATM.

There are no any breaking changes. Just possibility to use this package with newer versions of jest.

## Test plan

Not needed.

- [ ] Yes
- [X] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "github-files-fetcher": "^1.6.0",
     "glob": "^8.0.3",
     "husky": "^8.0.1",
-    "jest": "^29.1.2",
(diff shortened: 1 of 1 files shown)


ahnpnl:
Unfortunately we can’t do this. Doing like this will potentially break stuffs since Jest major version impacts this repo.

Issue #1774, opened 2022-09-15
[Bug]: npm install fails with latest jest 29.0.3


qwertyuiopzxcvbnmlkjhgfdsa1:
### Version

2.2.2

### Steps to reproduce

1,Use jest 29.0.3 as dependency with latest jest-preset-angular 2.2.2
2, npm install fails ERESOLVE unable to resolve dependency tree

### Expected behavior

npm install works

### Actual behavior

 npm install
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: xxx@3.0.1
npm ERR! Found: jest@29.0.3
npm ERR! node_modules/jest
npm ERR!   dev jest@"29.0.3" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer jest@"^28.0.0" from jest-preset-angular@12.2.2
npm ERR! node_modules/jest-preset-angular
npm ERR!   dev jest-preset-angular@"12.2.2" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.

### Additional context

Would you release jest-angular-preset with jest 29 supported?

### Environment

```shell
System:
    OS: Linux 5.18 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (12) x64 Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
  Binaries:
    Node: 16.17.0 - /usr/bin/node
    npm: 8.15.0 - /usr/bin/npm
```

ahnpnl:
We haven’t finished supporting Jest 29 yet so ye this will have to wait for a major release.

bendehghan:
Any updates/guesses on when 29 will be supported?

bendehghan:
Anything yet? It's been 2 weeks. Can you give an estimate please? At least we can plan things better on our end. Thanks

qwertyuiopzxcvbnmlkjhgfdsa1:
We downgraded Jest to version 28 and managed to get project working.

bendehghan:
Well, yeah, my project already works fine with Jest 28. We just wanted to keep up with Jest as it moves forward. I guess we'll wait until this is issue resolved and 29 is supported.

[...]

PR #1901, opened 2023-01-07
feat: support Jest 29

Tommy228:
## Summary

This PR is my attempt at addressing #1774 and supporting Jest 29.

This migrates all `ts-jest` configs under `globals`, which is deprecated, to transformers config.
 
It also exposes a new `defaultTransformerOptions` object to allow overriding the default transformer more easily.

```js
const { defaultTransformerOptions } = require('jest-preset-angular/presets');
```

## Test plan

I linked it to one of my biggest projects running Angular 15 with 3000+ unit tests. The tests have passed with no warning.

- [X] Yes

Overriding the config via globals may not work:
```js
module.exports = {
  globals: {
    'ts-jest': {
      tsconfig: '<rootDir>/src/tsconfig.spec.json',
    },
  },
};
```

Because the preset now provides the config via the transformer, the default configuration (like `tsconfig`) takes precedence over the globals configuration.

Instead, this needs to be done (unless there is a better solution?):
```js
const { defaultTransformerOptions } = require('jest-preset-angular/presets');

module.exports = {
  ...defaultConf,
  transform: {
    '^.+\\.(ts|js|mjs|html|svg)$': [
      'jest-preset-angular',
      {
        ...defaultTransformerOptions,
        tsconfig: '<rootDir>/src/tsconfig.spec.json',
      },
    ],
  },
};
```

## Other information

Not sure if I did this right. I'm struggling to get [NgJestTransformer](https://github.com/Tommy228/jest-preset-angular/blob/main/src/ng-jest-transformer.spec.ts) unit tests to work. They are working individually but they fail when ran together. I think it has to do with some caching stuff behind `TsJestTransformer`.

[@angular-builders/jest](https://github.com/just-jeb/angular-builders) needs to be updated as well because it passes the `tsconfig` via globals, which does not work anymore.

## Todo

- [x] Investigate failing `TsJestTransformer` tests
- [x] Upgrade `ts-jest` to include this [patch](https://github.com/kulshekhar/ts-jest/pull/3966)
- [ ] Create 13.x documentation (remove changes done to 12.x)
- [ ] Update examples

### DIFF
--- a/e2e/ast-transformers/hoisting/jest-isolated.config.js
+++ b/e2e/ast-transformers/hoisting/jest-isolated.config.js
@@ -2,9 +2,13 @@ const baseCfg = require('./jest.config');
 
 module.exports = {
   ...baseCfg,
-  globals: {
-    'ts-jest': {
-      isolatedModules: true,
-    },
+  transform: {
+    ...baseCfg.transform,
+    '^.+\\.(ts|js|html)$': [
+      '<rootDir>/../../../build/index.js',
+      {
+        isolatedModules: true,
+      },
+    ],
   },
 };
--- a/e2e/ast-transformers/ng-jit-transformers/jest-isolated.config.js
+++ b/e2e/ast-transformers/ng-jit-transformers/jest-isolated.config.js
@@ -2,10 +2,14 @@ const baseCfg = require('./jest.config');
 
 module.exports = {
   ...baseCfg,
-  globals: {
-    'ts-jest': {
-      ...baseCfg.globals['ts-jest'],
-      isolatedModules: true,
-    },
+  transform: {
+    ...baseCfg.transform,
+    '^.+\\.(ts|js|mjs|html)$': [
+      '<rootDir>/../../../build/index.js',
+      {
+        ...require('./ts-jest.config'),
+        isolatedModules: true,
+      },
+    ],
   },
 };
--- a/e2e/ast-transformers/ng-jit-transformers/jest.config.js
+++ b/e2e/ast-transformers/ng-jit-transformers/jest.config.js
@@ -1,14 +1,9 @@
 module.exports = {
-  globals: {
-    'ts-jest': {
-      stringifyContentPathRegex: '\\.html$',
-    },
-  },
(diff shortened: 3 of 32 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,36 @@
-## [12.2.3](https://github.com/thymikee/jest-preset-angular/compare/v12.2.2...v12.2.3) (2022-11-26)
+## [13.0.0](https://github.com/thymikee/jest-preset-angular/compare/v13.2.3...v13.0.0) (2023-01-07)
```
ahnpnl, line 1: You don't need to update changelog because it is generated via conventional changelog later

ahnpnl (commented):
so far it's in good direction. Using Jest transform option like this is the correct way

Tommy228:
Thanks for having a look! 

After some debugging, I think I found what's causing my tests to fail in sequence. 

It was indeed a caching issue on `ts-jest` side. `TsJestTransformer` uses the `globals` config as cache key [before it is merged with the transform config](https://github.com/kulshekhar/ts-jest/blob/main/src/legacy/ts-jest-transformer.ts#L90). So the tests use the cached configuration, even if transform options have changed. I'll create a PR to fix this.

Tommy228:
Small update about the current status: I won't have time to continue working on this before the end of the week. But what's left to do (cf. the tasks):
* Upgrade ts-jest to include this [patch](https://github.com/kulshekhar/ts-jest/pull/3966) (but this requires a release of `ts-jest`)
* Create a 13.x documentation (remove changes done to 12.x)
* Update the examples under `/examples` to use transform
* And of course, test that this didn't break anything

Help is welcome

viceice:
Your patch was released as v29.0.4🎉 

- https://github.com/kulshekhar/ts-jest/blob/HEAD/CHANGELOG.md#%E2%80%8B2904-httpsgithubcomkulshekharts-jestcomparev2903v2904-2023-01-10

kbrilla:
@Tommy228 hey any updates on this? Is it ready for the second round of review maybe?

ahnpnl:
done in #1937

PR #2007, opened 2023-03-16
Fix angular version

ajsaraujo:
Fixed the README. The text said "Angular < 13" but the article is actually about Angular >= 13, the opposite.


ahnpnl:
It is the guide for angular 12 or before to go to 13+

Issue #2162, opened 2023-09-12
[Bug]: latest zone.js v0.13.2 is causing import error


ghiscoding:
### Version

13.1.1

### Steps to reproduce

1. clone my repo at https://github.com/ghiscoding/Angular-Slickgrid/
2. bump `zone.js` to `v0.13.2`
3. `yarn install # yarn classic`
4. `yarn jest:coverage # or yarn jest:watch`

### Expected behavior

No error thrown

### Actual behavior

I'm getting the following error (which can be seen under this GitHub Action [PR failure](https://github.com/ghiscoding/Angular-Slickgrid/actions/runs/6153316033/job/16697173514?pr=1250))

```sh
$ jest --runInBand --coverage=true --ci --config ./test/jest.config.ts
FAIL src/app/modules/angular-slickgrid/components/__tests__/angular-slickgrid.component.spec.ts
  ● Test suite failed to run

    Cannot find module 'zone.js/bundles/zone-testing-bundle.umd' from 'node_modules/jest-preset-angular/setup-jest.js'

    Require stack:
      node_modules/jest-preset-angular/setup-jest.js
      test/setup-jest.ts

    > 1 | import 'jest-preset-angular/setup-jest';
        | ^
      2 | import './jest-global-mocks';

      at Resolver._throwModNotFoundError (node_modules/jest-resolve/build/resolver.js:42[7](https://github.com/ghiscoding/Angular-Slickgrid/actions/runs/6153316033/job/16697173514?pr=1250#step:10:8):[11](https://github.com/ghiscoding/Angular-Slickgrid/actions/runs/6153316033/job/16697173514?pr=1250#step:10:12))
      at Object.<anonymous> (node_modules/jest-preset-angular/setup-jest.js:1:1)
      at Object.<anonymous> (test/setup-jest.ts:1:1)
```

### Additional context

The error came from Renovate which runs once a week in my project, the dependency `zone.js` was updated less than a day ago and the import issue seem to come from this Angular [PR](https://github.com/angular/angular/pull/51652) and I'm not sure if the problem occurs in Angular or in Jest-Preset-Angular but I think it's latter and some imports might need to be updated as per the Angular [PR](https://github.com/angular/angular/pull/51652)

[...]

PR #2163, opened 2023-09-13
refactor: import zone.js/testing instead of zone-testing-bundle

JiaLiPassion:
Close https://github.com/thymikee/jest-preset-angular/issues/2162

From zone.js 0.13.2, the original
`zone.js/bundles/zone-testing-bundle.js` is not easily import and will eventually removed from zone.js 0.14.0.

The proper way to import zone.js and zone.js/testing is

```
import 'zone.js';
import 'zone.js/testing';
```

So this PR use this way to import zone.js related bundles.

### DIFF
--- a/setup-jest.js
+++ b/setup-jest.js
@@ -1,4 +1,5 @@
-require('zone.js/bundles/zone-testing-bundle.umd');
+require('zone.js');
+require('zone.js/testing');
 const { getTestBed } = require('@angular/core/testing');
 const {
   BrowserDynamicTestingModule,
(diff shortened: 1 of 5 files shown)


alan-agius4 (approved):
LGTM

ocombe:
@ahnpnl could you make a new release with that fix please?

ghiscoding:
@ahnpnl we really need a patch release for this since as pointed out by this [comment](https://github.com/thymikee/jest-preset-angular/issues/2162#issuecomment-1716244659), the temp fix that was introduced in Angular `zone.js` was reverted in `zone.js@0.14.0` and now the issue #2162 came back and is again a problem

> `zone.js` version `0.13.3` has been just released which should get around this issue.
> Deep imports will eventually no longer be allowed from `zone.js` version `0.14.0`.

PR #2164, opened 2023-09-15
fix: generated id="root{n}" should be removed

wellwind:
## Summary

When using snapshot testing, Angular will generate `<div id="root{n}">` root tag every time call `TestBed.createComponent`; this attribute will cause the snapshot not to be stable and should be removed when serializing content. 

## Test plan

I have updated the testing snapshot content, fixed the code, and the snapshot testing passed.

- [x] Yes
- [ ] No

All old snapshots containing `<div id="root{n}">` will cause failure.

So developers have to update the new snapshot manually after the upgrade.

### DIFF
--- a/e2e/snapshot-serializers/__tests__/__snapshots__/foo.component.spec.ts.snap
+++ b/e2e/snapshot-serializers/__tests__/__snapshots__/foo.component.spec.ts.snap
@@ -9,7 +9,9 @@ exports[`snapshot should work 1`] = `
 >
   <p>
     Line 1
-  </p><div>
+  </p><div
+    id="root0"
+  >
     <div>
        val1 
     </div>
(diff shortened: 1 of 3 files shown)


wellwind:
Hi @ahnpnl, 

do you think this PR is fine to approve?

e2e/snapshot-serializers/__tests__/__snapshots__/foo.component.spec.ts.snap
```diff
@@ -9,7 +9,9 @@ exports[`snapshot should work 1`] = `
 >
   <p>
     Line 1
-  </p><div>
+  </p><div
```
ahnpnl, line 12: can we avoid not generating this`id`?

ahnpnl:
Continue in #2565

Issue #2138, opened 2023-07-10
[Bug]: ESM with TypeScript >= 5.0 hangs forever


dgp1130:
### Version

13.1.1 (current latest)

### Steps to reproduce

Clone https://github.com/dgp1130/ng-large/tree/jest-preset-angular-esm-bug and run:

```shell
npm ci
npm test # Runs `jest`
```

There's nothing particularly special about this repo. It's basically a brand new `ng new` app on latest Angular where I have attempted to set up `jest-preset-angular` using ESM.

### Expected behavior

Test should pass as there is only one no-op component in the repository.

### Actual behavior

Jest prints:

```
RUNS src/app/app.component.spec.ts
```

And then hangs there forever, never making any meaningful progress.

### Additional context

I'm using the current latest Angular, Jest, Node, and `jest-preset-angular`. Bisecting the versions a bit, I discovered that TypeScript 5.0 introduced the error. TS 4.9.4 does _not_ reproduce, while everything greater than 5.0 (up through 5.1.4, the current latest) fails.

While Jest hangs without an error message, if you open a debugger with `--expose-internals` (`npm run test:debug` in that repo), I was able to find the [failing assertion](https://github.com/microsoft/TypeScript/blob/69d4d57b61b3a5dcc3f9d172c11e0827a47f24fc/src/compiler/debug.ts#L196) which includes the error:

> Unhandled SyntaxKind: ImportClause.

I'm trying to use native ESM here, so I would think TypeScript should emit any imports as effectively a no-op and let Node handle the rest, so I'm not sure why TypeScript is failing to emit the import.

### Environment

```shell
OS: Debian Linux
Node: v18.16.1
Jest: v29.6.1
```

NicBright:
I've stumbled into this as well. Maybe my findings can help to find the underlying issue:

Part of the problem is related to `jest-preset-angular`'s `replace-resources.js` adding a statement, presumably because it wants to add an import for the `templateUrl: '...'` part of the `@Component()` decorator

Here are my findings when debugging via `inspect-brk` and attaching to the debug session with chrome's `about:inspect`:

(1) see: it's the same error message that @dgp1130 [observed](https://github.com/thymikee/jest-preset-angular/issues/2138#issue-1797754592)


![image](https://github.com/thymikee/jest-preset-angular/assets/1225659/3806c545-8b41-4120-a049-ab13292c23f9)

[...]

PR #2185, opened 2023-11-08
build: update project to Angular 17

michaelfaith:
## Summary

This change updates root dependencies to use Angular 17.  It also introduces a new example project with a fully standalone v17 app for testing.

Note: updating to the latest prettier / eslint resulted in some minor formatting updates.

- [ ] Yes
- [x] No

## Other information
The `esm` and `esm-isolated` tests under the new example 17 project were timing out for me, but I wasn't sure if that was just something wrong with my local environment, since I've not run this project locally before.  It would be good to confirm that these are running successfully.  If not, I may need some guidance on what might be the issue.  Might be related to this issue #2138


michaelfaith:
@ahnpnl Curious how you'd like to deal with this kind of code.  If we upgrade TS in the root to 5.2, several props don't exist anymore.  Would you prefer to keep the root version of TS lower to support this compatibility layer, or take it out and increase min version support?

image

ahnpnl:
> @ahnpnl Curious how you'd like to deal with this kind of code. If we upgrade TS in the root to 5.2, several props used in this code don't exist anymore. Would you prefer to keep the root version of TS lower to support this compatibility layer, or take it out and increase min version support?
> 
> image

The files in `src/transformers` and `src/ngtsc` are copy version of Angular codes (with a bit modifications, mostly in `replace-resource.ts`), which are downloaded via `https://github.com/thymikee/jest-preset-angular/blob/main/scripts/prebuild.js`.

According to Angular support policy, after **22th Nov 2023**, the minimum supported version is **v15**. That means the files in The files in `src/transformers` and `src/ngtsc` need to be able to run with Angular 15, 16 and 17.

If some codes have been removed in **v17**, but still exists in **v15** and **v16**, we need to:
- First download the changes in **v17**
- Compare with downloaded version of **v15** and **v16**
- Make the changes in such a way to support all.

michaelfaith:
Great. thanks for the background.  I'll look at the new versions of these files

johnc-ftl:
Thank you for your work on this. I haven't grabbed this branch, but I would bet money that the tests are timing out due to the issue described in #2138. I have not been able to upgrade to Typescript 5 due to #2138. And ng 17 requires TS 5, so that issue definitely needs a fix for ng 17 to work.

johncrim:
Thank you for your work on this. I haven't grabbed this branch, but I would bet money that the tests are timing out due to the issue described in #2138. I have not been able to upgrade to Typescript 5 due to #2138. And ng 17 requires TS 5, so that issue definitely needs a fix for ng 17 to work.

michaelfaith:
Yeah, I didn't anticipate the TypeScript compatibility work that's ultimately required to land this Angular upgrade.  Realistically, it'll be at least another couple weeks before I can get back on this and start to work through that refactor.  Since the TypeScript upgrade could be done separately, if anyone else wants to jump in and do that as a separate change, please feel free.

ahnpnl:
Continue in #2198

michaelfaith:
Thanks for wrapping this up.  Apologies for the delay on my end.  We had a long holiday.

Issue #2196, opened 2023-12-20
[Feature]: Angular 17 support 


devgojokotech:
### 🚀 Feature Proposal

We need to update the import `zone.js/bundles/zone-testing.js` and `zone.js/dist/zone` to `zone.js` and `zone.js/testing` as they are updates now in angular version 17, it leads to incompatible issue while upgrading from 17.

### Motivation

It will add compatibility for angular 17.

### Example

.

kpeters-cbsi:
So is `jest-preset-angular` not compatible with Angular 17?

devgojokotech:
Yeah its not as of now, due to zonejs import path
<img width="867" alt="Screenshot 2023-12-21 at 10 37 42 AM" src="https://github.com/thymikee/jest-preset-angular/assets/118733747/5674a328-8ee9-469f-a594-492d704ffee2">

cfrezier:
We also need this.

ahnpnl:
We already made a release with updated zone import, see https://github.com/thymikee/jest-preset-angular/pull/2163

ahnpnl:
Should be fixed with **v13.1.5**

PR #2197, opened 2024-01-03
Fix TS5 + ESM hang; and upgrade dev dependencies to ng17

### LINKED ISSUE #2138: [Bug]: ESM with TypeScript >= 5.0 hangs forever

johncrim:
## Summary

I attempted to fix #2138, and as a side effect also upgrade to ng 17, resolving #2196. The main fixes were adding/improving the typescript version detection and version bridging functions so that the breaking changes in TS 5 (that I'm aware of) are handled, without changing the current level of backward compatibility with older versions of typescript and angular.

## Test plan

`yarn test` and `yarn test-examples` all pass (locally).

Tomorrow I will upgrade our projects to ng 17 and TS 5 to verify that our tests no longer hang, as described in #2138.

- [ ] Yes
- [x] No

Should not be breaking - I did my best to preserve the existing level of backward compatibility.

## Other information

fixes: #2138 
resolves: #2196

### DIFF
--- a/src/transformers/replace-resources.ts
+++ b/src/transformers/replace-resources.ts
@@ -9,6 +9,7 @@ import type { TsCompilerInstance } from 'ts-jest';
 import ts from 'typescript';
 
 import { STYLES, STYLE_URLS, TEMPLATE_URL, TEMPLATE, REQUIRE, COMPONENT, STYLE_URL } from '../constants';
+import { getDecorators, getModifiers, toMutableArray } from '../ngtsc/ts_compatibility';
 
 const isAfterVersion = (targetMajor: number, targetMinor: number): boolean => {
   const [major, minor] = ts.versionMajorMinor.split('.').map((part) => parseInt(part));
@@ -23,6 +24,9 @@ const isAfterVersion = (targetMajor: number, targetMinor: number): boolean => {
 };
 
 const IS_TS_48 = isAfterVersion(4, 8);
+
+/** Whether the current TypeScript version is after 5.0. */
+const IS_AFTER_TS_50 = isAfterVersion(5, 0);
 
 const shouldTransform = (fileName: string) => !fileName.endsWith('.ngfactory.ts') && !fileName.endsWith('.ngstyle.ts');
 /**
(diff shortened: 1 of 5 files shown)


src/transformers/replace-resources.ts
```diff
@@ -136,7 +140,7 @@ function visitClassDeclaration(
         node.heritageClauses,
         node.members,
       )
-    : nodeFactory.updateClassDeclaration(
+    : (nodeFactory as any).updateClassDeclaration(
```
ahnpnl, line 143: Are there other ways we can avoid casting `any`?

johncrim:
@ahnpnl - I had to add `husky` as a dev dependency - the yarn cache build steps were failing b/c it wasn't there. For some reason it wasn't on the `main` that I branched from.

Thanks for the review + guidance.

ahnpnl:
Would you please also add eslint comments to bypass several `Warning:   143:23  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any`, make it look prettier 😃

johncrim:
I can confirm that these updates work as expected on our projects using TS 5.2.2 and ng 17.0.8 and ng 17.0.9.

ahnpnl (approved):

ahnpnl:
Thanks!

johncrim:
Happy to help! Thank you!

michaelfaith:
Great work on this change!

Issue #2366, opened 2024-04-02
[Bug]: jest-global-mocks should set the value of `CSS` to undefined and not null


straker:
### Version

latest

### Steps to reproduce

Use the [jest-gobal-mocks](https://thymikee.github.io/jest-preset-angular/docs/getting-started/installation/#global-mocks) file in the angular project and try to load a 3rd party library that use `window.CSS`. Example https://colorjs.io/

### Expected behavior

Runs without issues

### Actual behavior

Errors with 

```
Uncaught TypeError: Cannot read properties of null (reading 'supports')
```

### Additional context

`null` is an incorrect override for the `CSS` property as it should be `undefined` by default in browsers that do not support it. Setting it to `null` causes issues with 3rd party libraries that guard against using it and expect `undefined`. Example: [colorjs.io](https://github.com/color-js/color.js/pull/472).

### Environment

```shell
System:
    OS: macOS 14.2.1
    CPU: (12) arm64 Apple M2 Pro
  Binaries:
    Node: 18.17.1 - ~/.nvm/versions/node/v18.17.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.17.1/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.1/bin/npm
```

PR #2367, opened 2024-04-02
docs: CSS module should be set to undefined

michael-siek:
closes: https://github.com/thymikee/jest-preset-angular/issues/2366

### DIFF
--- a/examples/example-app-monorepo/apps/app1/jest-global-mocks.ts
+++ b/examples/example-app-monorepo/apps/app1/jest-global-mocks.ts
@@ -1,7 +1,5 @@
 import { jest } from '@jest/globals';
 
-Object.defineProperty(window, 'CSS', { value: null });
-
 Object.defineProperty(document, 'doctype', {
   value: '<!DOCTYPE html>',
 });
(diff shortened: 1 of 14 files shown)


dbjorge (changes requested):
Several `jest-global-mocks.ts` files within projects under `/examples` also feature a null-CSS mock line similar to the one removed from the doc here. @michael-siek, we should update those examples in line with the doc update.

@ahnpnl , those examples already appear to be run as e2e tests within the repo's CI builds; do you agree that updating them would be sufficient e2e testing for this change, or are you looking for something new?

ahnpnl:
Yes I would say just need to update an existing test in `examples` folder should be sufficient. 

For this common test, let's try to add it to all example apps under `examples` folder. We try to keep all the tests in those example apps identical to make it easy to migrate to a new major version.

ahnpnl:
Only small thing, would you please also update similar files at https://github.com/thymikee/jest-preset-angular/tree/main/website/versioned_docs? We have several versioned docs which need to be updated too

ahnpnl (approved):

PR #2369, opened 2024-04-04
Add extra configuration on jest.setup.js to support Ionic 6 or 7

orl99:
## Summary

This PR adds how to fix one issue when we are having Ionic 6 or 7 Framework in the project

We have to include this important configuration in the documentation so that all who are using Ionic 6 or 7 will avoid houuurs of investigation.

## Test plan

Error: 
image

Solution:

```js tab
module.exports = {
  // ...other options
transformIgnorePatterns: ['<rootDir>/node_modules/(?!(@ionic/core|@ionic/angular|@stencil/core|.*\\.mjs$))']
};
```

image

- [ ] Yes
- [ x ] No

### DIFF
--- a/website/docs/guides/angular-13+.md
+++ b/website/docs/guides/angular-13+.md
@@ -172,3 +172,25 @@ const jestConfig: Config = {
 
 export default jestConfig;
 ```
+
+### Usage with Ionic 6 or 7
+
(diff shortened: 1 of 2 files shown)


ahnpnl:
To fix CI, you need to run `lint-prettier` script at root `package.json`

orl99:
sure, let me do it

PR #2462, opened 2024-05-20
feat: extend version support range to Angular 18

leosvelperez:
## Summary

Angular 18 should be [released sometime this week](https://angular.dev/reference/releases#release-schedule) (the Week of November 20th, 2024). This PR allows users to install `jest-preset-angular` without getting `peerDependency` errors, which cause some package managers to fail the installation process.

## Test plan

We've tested `jest-preset-angular` + Angular 18 as part of [the Angular 18 support for Nx](https://github.com/nrwl/nx/pull/22509), and everything works other than a peer dependency error.

Additionally, I created a separate PR with a new example test application using the latest RC version of Angular 18: https://github.com/thymikee/jest-preset-angular/pull/2463. The CI checks are all passing. I'll update that PR when the stable version of Angular 18 is released.

- [ ] Yes
- [X] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,10 +55,10 @@
     "esbuild": ">=0.15.13"
   },
   "peerDependencies": {
-    "@angular-devkit/build-angular": ">=15.0.0 <18.0.0",
-    "@angular/compiler-cli": ">=15.0.0 <18.0.0",
-    "@angular/core": ">=15.0.0 <18.0.0",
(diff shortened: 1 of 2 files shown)


leosvelperez:
Thank you! Do you think you could release the change? It would allow us at Nx to support the new version before the Angular release this week. Given that this PR doesn't change any functionality, the change could be released before the Angular version is released.

ahnpnl:
just did 👍

leosvelperez:
Awesome! Thanks!

PR #2615, opened 2024-07-22
test: add unit tests for `transpileModule` with JIT transformers

ahnpnl:
## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information
**N.A.**


ahnpnl:
Hi @dgp1130, as discussed via DM before, I have copied most of tests from https://github.com/angular/angular/blob/main/packages/compiler-cli/src/ngtsc/transform/jit/test/downlevel_decorators_transform_spec.ts Some tests I found not needed for Jest so I didn't include them here. The amount of copied tests in this PR I think should be enough to cover all cases.

There are indeed a few cases where `transpileModule` doesn't work, which I mark them with `it.failing`. Would you please help me to take a look whenever there is time? Thank you!

src/compiler/ng-jest-compiler.spec.ts
```diff
@@ -1,34 +1,587 @@
...
+
+            expect(diagnostics.length).toBe(0);
+            expect(code).toContain('const other_file_1 = require("./other-file");');
+            expect(code).toContain(dedent`
+                MyDir.ctorParameters = () => [
+                    { type: other_file_1.MyOtherClass }
+                ];
+                exports.MyDir = MyDir = tslib_1.__decorate([
+                    (0, core_1.Directive)(),
+                    tslib_1.__metadata("design:paramtypes", Object])
```
dgp1130, line 284: My understanding of `emitDecoratorMetadata` is that it changes the output slightly based on whether or not the constructor parameter type is local or imported. It wants to emit `tslib_1.__metadata("design:paramtypes", MyOtherClass)`, but TS can't introspect the type of an imported symbol with `isolatedModules` and doesn't know `MyOtherClass` is a `class`. Instead, TS needs to generate a small JS snippet to determine this at runtime. See the output from [this example](https://www.typescriptlang.org/play/?experimentalDecorators=true&emitDecoratorMetadata=true&target=10&jsx=0&module=7&isolatedModules=true&isolatedDeclarations=true#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwGEcBbABzxFQwAoBKALnmo1gHMQNGpUBPW+ALwA+eADccWYAG4AUDID08+AAEMAZwC0WVqhxwZWMnozwA3gEkjMDCGABlEDFFYwIAL5IYJeAHIAdPJqjs6uPrIykFBqavAAMjhgUBAOTi4Ipm5yykRGqJQ0tDIgAB7k1vCR0fAAsjw55HlUZjLwrRV4ahgwyJh61C1tg6QwWKJQNvBwUMB4EDzwEAlJjPGJycFpADQDg63Do+MIUzOoc-CGZTbAjJaXtikhINuD-BkymUA). Is that the output you get which causes the test failure?

Interestingly, `tsc` seems to add this snippet whether or not `isolatedModules` is enabled. In theory I can see that TS wouldn't know `MyOtherClass` is a class as opposed to an interface and maybe needs to pick `Object` because it can't be certain `MyOtherClass` is valid in a value reference position.

src/compiler/ng-jest-compiler.spec.ts
```diff
@@ -1,34 +1,587 @@
...
+                exports.MyComp = MyComp = tslib_1.__decorate([
+                    (0, core_1.Component)({ template: 'hello' })
+                ], MyComp);
+            `);
+        });
+
+        it.failing(
+            'should allow for type-only references to be removed with `emitDecoratorMetadata` from custom decorators',
+            () => {
+                // This test only passes when using `import type { ExternalInterface } from './external-interface';`
```
dgp1130, line 565: This one makes sense to me because with `isolatedModules` TS shouldn't know that it's an interface and not eligible to be moved into a value reference position.

src/compiler/ng-jest-compiler.spec.ts
```diff
@@ -1,34 +1,587 @@
...
+                    (0, core_1.Directive)()
+               ], MyDir);
+            `);
+        });
+
+        it.failing('should properly serialize constructor parameter with external qualified name type', () => {
+            // This test doesn't pass because `transpileModule` can't resolve `import *`
+            const { code, diagnostics } = transformCjs(`
+                import {Directive} from '@angular/core';
+                import * as externalFile from './other-file';
```
dgp1130, line 319: I'm surprised this would be different than `import { MyOtherClass } from './other-file';` and makes me wonder if there's a bug in [`entityNameToExpression`](https://github.com/angular/angular/blob/5dcdbfcba934a930468aec140a7183b034466bdf/packages/compiler-cli/src/transformers/jit_transforms/downlevel_decorators_transform.ts#L351) somewhere? We have special handling of aliased imports, I wonder if we also need special handling for namespaced imports?

src/compiler/ng-jest-compiler.spec.ts
```diff
@@ -1,34 +1,587 @@
...
+                    { type: other.OtherClass }
+               ];
+               exports.MyDir = MyDir = tslib_1.__decorate([
+                    (0, core_1.Directive)()
+               ], MyDir);
+            `);
+        });
+
+        it.failing('should properly downlevel constructor parameter decorators', () => {
+            // This test seems to not pass because `Document` type is unknown with `transpileModule`
```
dgp1130, line 366: What if you included in the `lib.d.ts` files with `ts.transpileModule`? Would that bring `Document` into scope?

I don't see any reason TypeScript _couldn't_ know about this and still get all the isolation benefits.

src/compiler/ng-jest-compiler.spec.ts
```diff
@@ -1,34 +1,587 @@
...
+            `);
+
+            expect(diagnostics.length).toBe(0);
+            expect(code).not.toContain('external');
+            expect(code).toContain(dedent`
+                MyDir.ctorParameters = () => [
+                    { type: undefined, decorators: [{ type: core_1.Inject, args: ['$state',] }] },
+                    { type: undefined, decorators: [{ type: core_1.Inject, args: ['$overlay',] }] },
+                    { type: undefined, decorators: [{ type: core_1.Inject, args: ['$default',] }] },
+                    { type: undefined, decorators: [{ type: core_1.Inject, args: ['$keyCodes',] }] }
```
dgp1130, line 531: If `type` is expected to be `undefined`, then why are we trying to resolve the `import *`? It doesn't seem like we shouldn't need that type? I wonder if the correct behavior on resolution error would just be to fall back to `any` as the type (and emit `undefined`)?

dgp1130 (commented):
Left some general thoughts on the failing tests. I don't know that I have great answers for why they're failing, but hopefully the extra context is helpful.

I do wonder how much of this is `downlevel_decorators_transform` being incompatible with `isolatedModules` and how much is just not taking the effort to support it. Would some of these failures be fixable upstream or at least gracefully degrade and/or give a better error message about what works?

src/compiler/ng-jest-compiler.spec.ts
ahnpnl: Thanks for the hint! I was missing this when reading source code of `transpileModule` and indeed it helped 😃

src/compiler/ng-jest-compiler.spec.ts
ahnpnl: Yes the output from TS playground was exactly the output which caused the test failture. If we leave the output codes to be similar like TS playground, will that cause any issues when testing real application codes? 

Is `emitDecoratorMetadata: true` still important to Angular apps these days? I remember somewhere I saw a commit from Angular which allows to disable `emitDecoratorMetadata` and application still works.

src/compiler/ng-jest-compiler.spec.ts
ahnpnl: This case produces very interesting output
```
MyDir.ctorParameters = () => [
{ type: undefined, decorators: [{ type: core_1.Inject, args: ['$state',] }] },
{ type: external_1.IOverlay, decorators: [{ type: core_1.Inject, args: ['$overlay',] }] },
{ type: external_2.default, decorators: [{ type: core_1.Inject, args: ['$default',] }] },
{ type: external_1.KeyCodes, decorators: [{ type: core_1.Inject, args: ['$keyCodes',] }] }
];
exports.MyDir = MyDir = tslib_1.__decorate([
(0, core_1.Directive)()
], MyDir);
```

src/compiler/ng-jest-compiler.spec.ts
ahnpnl: I can confirm import name and import default work, only namespace import doesn't

src/compiler/ng-jest-compiler.spec.ts
dgp1130: I think that emitted JS is probably fine? Hard to be 100% sure unfortunately.

`ng new` does not enable `emitDecoratorMetadata` in the scaffolded `tsconfig.json` files, so I suspect it's not required, but I wouldn't be surprised if there are other effects from enabling it. I can ask around and see if anyone has more context.

src/compiler/ng-jest-compiler.spec.ts
dgp1130: Hmm, any of these could be an interface which wouldn't be valid in `type` though. I wonder if the type system is falling back to `any` because it can't be resolved and then we're assuming the import is a valid value reference. I wonder if the safer assumption would be to fall back to `type: undefined` in that scenario?

src/compiler/ng-jest-compiler.spec.ts
dgp1130: `typeof external_interface_1.ExternalInterface` sounds wrong to me. Interfaces are erased by TypeScript, so this expression can't work at runtime. I would expect this to be something like `Object` in this scenario since it can't be any more specific about the type.

I don't think Angular uses `design:returntype` anywhere. I think we only care about constructor parameters. So it probably doesn't matter what's actually in the `__metadata` call, as long as it doesn't throw.

src/compiler/ng-jest-compiler.spec.ts
ahnpnl: Since I’m using a “standalone” `Program` which doesn’t have enough information to validate which imports are interfaces or non interfaces, as well as not able to validate which import paths can be resolved which leads to the situation where the `Program` just assumes those paths exist or those are non interfaces?

ahnpnl: Here are a few things from my debugging:

### Failed scenarios

- Given file content:
```
import {Directive} from '@angular/core';
import externalFile from './other-file';
import * as foo from './other-file-foo';

@Directive()
export class MyDir {
      constructor(other: externalFile.MyOtherClass,
                           hello: foo.MyFooClass) {}
}
```

- `SymbolObject` shape:
```
symbol SymbolObject {
      flags: 34078720,
      escapedName: 'MyOtherClass',
      declarations: undefined,
      valueDeclaration: undefined,
      id: 0,
      mergeId: 0,
      parent: SymbolObject {
        flags: 34078720,
        escapedName: 'externalFile',
        declarations: undefined,
        valueDeclaration: undefined,
        id: 0,
        mergeId: 0,
        parent: undefined,
        members: undefined,
        exports: undefined,
        exportSymbol: undefined,
        constEnumOnlyModule: undefined,
        isReferenced: undefined,
        lastAssignmentPos: undefined,
        links: SymbolLinks { checkFlags: 1048576, declaredType: [TypeObject] }
      },
      members: undefined,
      exports: undefined,
      exportSymbol: undefined,
      constEnumOnlyModule: undefined,
      isReferenced: undefined,
      lastAssignmentPos: undefined,
      links: SymbolLinks {
        checkFlags: 1048576,
        declaredType: TypeObject {
          flags: 1,
          checker: [Object],
          id: 6,
          intrinsicName: 'unresolved',
          debugIntrinsicName: undefined,
          objectFlags: 52953088
        }
      }
 }
```
with this symbol object, I will always run into this line https://github.com/angular/angular/blob/d108320ea394e827fc23011b54d12e168287bce4/packages/compiler-cli/src/ngtsc/transform/jit/src/downlevel_decorators_transform.ts#L363

### Passed scenario

- Given file content
```
import {Directive} from '@angular/core';
import {MyOtherClass} from './other-file';

@Directive()
export class MyDir {
       constructor(other: MyOtherClass) {}
}
```

- `SymbolObject` shape:
```
symbol <ref *1> SymbolObject {
      flags: 2097152,
      escapedName: 'MyOtherClass',
      declarations: [
        NodeObject {
          pos: 82,
          end: 94,
          kind: 276,
          id: 0,
          flags: 0,
          modifierFlagsCache: 536870912,
          transformFlags: 0,
          parent: [NodeObject],
          original: undefined,
          emitNode: undefined,
          symbol: [Circular *1],
          localSymbol: undefined,
          isTypeOnly: false,
          propertyName: undefined,
          name: [IdentifierObject]
        }
      ],
      valueDeclaration: undefined,
      id: 9,
      mergeId: 0,
      parent: undefined,
      members: undefined,
      exports: undefined,
      exportSymbol: undefined,
      constEnumOnlyModule: undefined,
      isReferenced: 788968,
      lastAssignmentPos: undefined,
      links: undefined
}
```

with this `SymbolObject`, it will never run into this https://github.com/angular/angular/blob/main/packages/compiler-cli/src/ngtsc/transform/jit/src/downlevel_decorators_transform.ts#L363

dgp1130: In retrospect, `any` might not necessarily be a problem. Both type references and value references can be typed `any`, so that might be fine. I wonder if it's possible to ask TS whether or not each import is a value reference or a type reference. I'm guessing it's assuming imported symbols are a value reference when it can't resolve the import and we'd probably want to reduce that assumption to a type reference in this scenario.

Unfortunately I'm not very familiar with the TS API and docs are pretty lacking, so I'm not sure about the best way to verify that.

ahnpnl: For the 2nd case, can we unconditionally *NOT* generating `undefined` so that it’s possible to run tests and leave it up to Jest runtime to validate if that is valid runtime value? 

For example:

`import foo from ‘./foo’` would resolve in `type` as `type: foo.<something>`

And the same applies to `import * as foo from ‘./foo’`

If we set `type: undefined`, it will break all existing tests and it makes it not possible to run tests in transpilation mode because there will be error like `Can't resolve all parameters for AppComponent: (?)` which I observe when debugging this issue https://github.com/thymikee/jest-preset-angular/issues/963

Francesco-Borzi:
hi, we are experiencing quite some performance issues when running unit tests using `jest-preset-angular`. I've been reading here and there on the web and bumped into this PR.

I wanted to first of all thank for your precious contributions, and I also wanted to ask if this PR is supposed to make tests run faster and how much is the expected impact.

ahnpnl:
@Francesco-Borzi this PR opens up a possibility to improve both test speed and memory usage without breaking existing test suites.

At the moment, you can already use `isolatedModules: true` but there are caveats:

- Some scenarios won’t work, which are demonstrated in this PR.
- Ignore type checking. This can be easily enabled

Francesco-Borzi:
> @Francesco-Borzi this PR opens up a possibility to improve both test speed and memory usage without breaking existing test suites.
> 
> At the moment, you can already use `isolatedModules: true` but there are caveats:
> 
> * Some scenarios won’t work, which are demonstrated in this PR.
> * Ignore type checking. This can be easily enabled

hi @ahnpnl , apart from the new compilation errors, I don't seem to get any performance improvement using `isolatedModules: true`

ahnpnl:
Just to be sure, I mean https://github.com/thymikee/jest-preset-angular/blob/fd36bec6176dfa6470414ad80f9b91a441fdaf4a/examples/example-app-v18/jest-isolated.config.js#L13 Often this option is mistaken with tsconfig option

Francesco-Borzi:
@ahnpnl thanks for the clarification!

PR #2880, opened 2024-11-28
chore: 🤖 remove @angular-devkit/build-angular peerDep

danymarques:
## Summary

This update ensures compatibility for projects still using Angular 18 by removing the need to explicitly list the `@angular-devkit/build-angular` dependency. This is achieved by enabling the direct use of builders from `@angular/build`. 

- [ ] Yes
- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "esbuild": ">=0.15.13"
   },
   "peerDependencies": {
-    "@angular-devkit/build-angular": ">=15.0.0 <20.0.0",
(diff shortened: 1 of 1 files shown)


ahnpnl (approved):
I think in the past we used to use this dep but now the existing codes don't use it anymore. It's safe to remove. Thanks!

danymarques:
> I think in the past we used to use this dep but now the existing codes don't use it anymore. It's safe to remove. Thanks!

Yes, I checked beforehand if it was used or not 😄

PR #2904, opened 2024-12-19
feat: add custom `jsdom` env

ahnpnl:
## Summary

See commits

## Test plan

Green CI

- [ ] Yes
- [x] No

## Other information
**N.A.**


Delagen:
@ahnpnl Can we make `jest-environment-jsdom` package optional now? Because if we use custom environment provided by you package we only need to install `jsdom` package.

ahnpnl:
Ya I think `jest-environment-jsdom` can be an optional peer dep. However, marking it as optional now is like a breaking change. I would reserve that change for the next major version.

PR #3079, opened 2025-04-28
chore: remove dependency on `@angular/platform-browser-dynamic`

JeanMeche:
The `@angular/platform-browser` package can be used instead.


ahnpnl:
Thanks @JeanMeche for initiative, close as implemented in #3247

PR #3826, opened 2026-06-08
ci: add node 24 to CI workflows

ahnpnl:

- [ ] Yes
- [x] No

## Other information
N/A


.github/workflows/pr-codes-check.yml
CharlieM312: should the tests run on node 24 given that the prepare cache job does?

ahnpnl: yes they should

CharlieM312 (approved):

PR #3838, opened 2026-06-12
build(deps): update to eslint 10

CharlieM312:
## Summary

Updates to eslint 10

## Test plan

Testing code works properly and only thing that fails is the OSV scan 

- [ ] Yes
- [x] No


examples/example-app-v22/package.json
CharlieM312, line 34: Awaiting 3 day waiting period to update this to 16.2.0

CharlieM312:
@ahnpnl PR is ready for review

ahnpnl (commented):
I think this PR is doing too many things

- If we only update `eslint`, let's keep it to eslint deps only. 
- I see that we are setting `resolutions` which is not recommended.

I suggest that you split up this PR into multiple PRs, each PR solves one particular issue.

About OSV scanner, there is an upstream bug https://github.com/google/osv-scanner-action/issues/132

CharlieM312:
> I think this PR is doing too many things
> 
> * If we only update `eslint`, let's keep it to eslint deps only.
> * I see that we are setting `resolutions` which is not recommended.
> 
> I suggest that you split up this PR into multiple PRs, each PR solves one particular issue.
> 
> About OSV scanner, there is an upstream bug [google/osv-scanner-action#132](https://github.com/google/osv-scanner-action/issues/132)

This PR has been updated to now do the eslint deps - there are also other minor code changes to make the example apps actually run. All uses of resolutions have been removed

performance/package.json
ahnpnl, line 1: pls revert the change for `performance` folder as well as the changes for `examples` and `e2e` folders

CharlieM312: Done - I have stashed them if needed for a separate PR

ahnpnl (approved):
Thanks!

PR #3843, opened 2026-06-12
feat(angular): make the example apps run

CharlieM312:
## Summary

This PR makes the example apps run properly through updating the typescript configuration and making a few small changes
Solves TS5096 error

- [ ] Yes
- [x] No

### DIFF
--- a/examples/example-app-v20/src/app/demo/demo.ts
+++ b/examples/example-app-v20/src/app/demo/demo.ts
@@ -153,9 +153,12 @@ export class InputValueBinderDirective {
 
     @Output() valueChange: EventEmitter<unknown> = new EventEmitter();
 
-    @HostListener('input', ['$event.target.value'])
-    onInput(value: unknown) {
-        this.valueChange.emit(value);
+    @HostListener('input', ['$event'])
+    onInput(event: Event) {
+        const target = event.target as HTMLInputElement | null;
+        if (target) {
+            this.valueChange.emit(target.value);
+        }
     }
 }
 
--- a/examples/example-app-v20/tsconfig.json
+++ b/examples/example-app-v20/tsconfig.json
@@ -11,7 +11,6 @@
     "importHelpers": true,
     "target": "ES2022",
     "module": "preserve",
-    "allowImportingTsExtensions": true,
     "paths": {
       "@shared/*": ["./src/app/shared/*"]
     }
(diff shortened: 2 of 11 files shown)


CharlieM312:
@ahnpnl pr can be reviewed

performance/package.json
ahnpnl, line 1: Let’s revert this performance folder, it’s not a part of the PR scope

examples/example-app-v21/yarn.lock
ahnpnl, line 1: Let’s revert this lock file change, we don’t change any deps in package.json, also the other lock files

examples/example-app-v21/src/main.ts
```diff
@@ -3,4 +3,6 @@ import { bootstrapApplication } from '@angular/platform-browser';
 import { App } from './app/app';
 import { appConfig } from './app/app.config';
 
+import 'zone.js';
```
ahnpnl: You can learn from this line https://github.com/thymikee/jest-preset-angular/blob/c78152e5fe80b5089e5607224e790b784ad2ad66/examples/example-app-v20/angular.json#L20

examples/example-app-v20/src/app/demo/demo.ts
```diff
@@ -153,9 +153,12 @@ export class InputValueBinderDirective {
 
     @Output() valueChange: EventEmitter<unknown> = new EventEmitter();
 
-    @HostListener('input', ['$event.target.value'])
-    onInput(value: unknown) {
-        this.valueChange.emit(value);
+    @HostListener('input', ['$event'])
```
ahnpnl, line 156: Why do we need this change? This syntax and the previous syntax are technically the same. 

If it is required to change to your proposed change, sound like a breaking change due to Angular version?

CharlieM312: this change is a fix given that currently the yarn start command will not work across the example apps unless this change has been made

ahnpnl: I see, so this one is about this error
```
 [ERROR] TS2339: Property 'value' does not exist on type 'EventTarget'. [plugin angular-compiler]

    src/app/demo/demo.ts:156:43:
      156 │     @HostListener('input', ['$event.target.value'])

```
the PR context didn't mention that so it wasn't clear why the app doesn't start

examples/example-app-v20/tsconfig.json
```diff
@@ -11,7 +11,6 @@
     "importHelpers": true,
     "target": "ES2022",
     "module": "preserve",
-    "allowImportingTsExtensions": true,
```
ahnpnl, line 14: This is to solve the error 
```
 TS5096: Option 'allowImportingTsExtensions' can only be used when either 'noEmit' or 'emitDeclarationOnly' is set. [plugin angular-compiler]
```

which wasn't described in PR description either

ahnpnl (approved):

PR #3916, opened 2026-07-09
build(deps): update to use newer @angular/build package

CharlieM312:
## Summary

Updates to use newer @angular/build package given this is not done automatically by the bots
Kept all versions the same as those will be updated automatically
Also needed to add @types/express as the testing had issues with this

## Test plan

Tests pass

- [ ] Yes
- [x] No

### DIFF
--- a/examples/example-app-monorepo/package.json
+++ b/examples/example-app-monorepo/package.json
@@ -28,9 +28,10 @@
     "zone.js": "~0.16.0"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^21.2.18",
+    "@angular/build": "^21.2.18",
     "@angular/cli": "^21.2.18",
     "@angular/compiler-cli": "^21.2.18",
+    "@types/express": "5.0.6",
     "@types/jest": "^30.0.0",
     "@types/node": "^24.13.3",
     "esbuild-register": "^3.6.0",
--- a/examples/example-app-monorepo/yarn.lock
+++ b/examples/example-app-monorepo/yarn.lock
@@ -181,131 +181,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@angular-devkit/build-angular@npm:^21.2.18":
-  version: 21.2.18
-  resolution: "@angular-devkit/build-angular@npm:21.2.18"
-  dependencies:
-    "@ampproject/remapping": "npm:2.3.0"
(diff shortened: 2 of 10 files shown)


CharlieM312:
@ahnpnl PR ready for review

examples/example-app-monorepo/package.json
```diff
@@ -28,9 +28,10 @@
     "zone.js": "~0.16.0"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^21.2.18",
+    "@angular/build": "^21.2.18",
     "@angular/cli": "^21.2.18",
     "@angular/compiler-cli": "^21.2.17",
+    "@types/express": "5.0.6",
```
ahnpnl: this is strange, we didn't need this type before. Why do we need to add now?

examples/example-app-monorepo/apps/app1/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "baseUrl": "./",
+    "ignoreDeprecations": "6.0",
```
ahnpnl: we are using ts 5.9 here, do we need this?

examples/example-app-monorepo/apps/app1/tsconfig.app.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "rootDir": "./",
```
ahnpnl: I remember ts 6.0 requires this, but here we are using 5.9, do we need this?

examples/example-app-v20/tsconfig.app.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "rootDir": "./src",
```
ahnpnl: same, we are not using ts 6.0 here, do we need this?

examples/example-app-monorepo/package.json
CharlieM312: not entirely sure, as shown here it fixes this error on the job: https://github.com/thymikee/jest-preset-angular/actions/runs/29021385896/job/86129540142. Does no harm being added

ahnpnl (approved):

PR #3945, opened 2026-07-18
build(deps): update babel deps to 8.0.1 and node version to minimum 22

CharlieM312:
## Summary

Updating babel deps to version 8 and updated project to run on minimum Node 22.
This is because Node 20 is at EOL as explained in comments, however node 22, 24 are on LTS and node 26 is new current
Reviewed the babel migration guide [https://babeljs.io/docs/v8-migration](url)
Note although the bot does do dependency updates it has a 3 day waiting period and babel 8 has been released for over a month

## Test plan

Tests pass

- [x] Yes
- [ ] No

### DIFF
--- a/.github/workflows/pr-codes-check.yml
+++ b/.github/workflows/pr-codes-check.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node-version: [20.x, 22.x, 24.x]
+        node-version: [22.x, 24.x, 26.x]
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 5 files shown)


CharlieM312:
@ahnpnl PR ready for review

CharlieM312:
In accordance with the node release schedule, we should now use node 22, 24, 26 for testing.
I will update the test matrices to use node 26. Node 20 is now EOL
image

CharlieM312:
Node release website https://nodejs.org/en/about/previous-releases

CharlieM312:
closing and reopening to re-run checks

ahnpnl:
- We have a direct transitive dep on Babel 7 via Jest and https://github.com/jestjs/jest/blob/f49721c78e195558b40913977c9230f5b7f559d8/packages/jest-transform/package.json#L22
- We still support Node 20 because Angular 20 is still using it hence it shouldn’t be removed. Node 20 will be dropped once Angular 20 is removed from the supported list.
- We depend on Jest a lot more than Angular. We only use the AST transformer source to bundle a compiled version.

With these deps, we can’t go for Babel 8. Even though angular CLI packages enforce Babel 8, as long as the build package still works, Babel 7 is still usable.

PR #3987, opened 2026-07-25
build(deps): bump react and docusaurus dependencies in /website

CharlieM312:
## Summary

Updating react and docusaurus dependencies in /website
This needs to be done occasionally as is not automatically done by renovate (as shown in #3837 )

## Test plan

Tests pass, website builds correctly locally

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #3991, opened 2026-07-25
fix: renovate file change to update docusaurus dependencies correctly

CharlieM312:
## Summary

Change to the renovate.json to update the docusaurus dependencies correctly and not leave out that in dependencies, as currently based on previous bumps as discussed in #3987 "devDependencies" are updated in /website just not the "dependencies" section.
Also tells renovate to update the caniuse-lite and baseline-browser-mapping packages to prevent this warning occuring when you build/start the example apps
image

## Test plan

Tests pass and if renovate says that this file is okay, on the next docusaurus version release it should also bump @docusaurus/core and other related dependencies

- [ ] Yes
- [x] No

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -43,6 +43,11 @@
       "rangeStrategy": "in-range-only"
     },
     {
+      "matchPaths": ["website/package.json"],
+      "matchPackagePatterns": ["^@docusaurus/", "^docusaurus"],
+      "rangeStrategy": "bump"
+    },
+    {
       "matchDepTypes": ["peerDependencies"],
       "enabled": false
     },


CharlieM312:
@ahnpnl small PR to address #3987

renovate.json
```diff
@@ -64,6 +68,10 @@
       "matchPaths": ["examples/**/package.json"],
       "matchDepNames": ["angular-in-memory-web-api", "zone.js"],
       "enabled": false
+    },
+    {
+      "matchPackageNames": ["caniuse-lite", "baseline-browser-mapping"],
+      "schedule": ["before 4am on Monday"]
```
CharlieM312: this can be changed depending on what's needed but scheduling may be good here given that this is updated regularly

ahnpnl: These are transitive deps which don’t exist in `package.json`. Setting this makes it hard to maintain since they are not obviously visible. I would prefer not adding these.

renovate.json
```diff
@@ -42,6 +42,10 @@
       "matchDepTypes": ["dependencies", "optionalDependencies"],
       "rangeStrategy": "in-range-only"
     },
+    {
+      "matchPackagePatterns": ["^?@docusaurus"],
```
ahnpnl: Why don’t we go with simpler one like `"matchPackagePatterns": ["^@docusaurus/", “^docusaurus”]`. This will cover all core deps of Docusaurus as well as any 3rd party plugins that start with `docusaurus` like https://github.com/thymikee/jest-preset-angular/blob/57b7a1a88a4b52c1edd091e8ca3a7a5aeba1e8ba/website/package.json#L23

Besides, we can add extra       `"matchPaths": ["website/package.json"]` to speed up scanning

ahnpnl (approved):
Thanks!

PR #4044, opened 2026-08-18
build(deps): bump typescript to latest non-majors

CharlieM312:
## Summary

bumped typescript to the latest non-major versions (this will be a one-time update given these are all the example apps using older versions of Angular).
Also updated jsdom version in /examples/example-app-monorepo to match /examples/example-app-v21 given these use the same version of Angular

## Test plan

Tests pass

- [ ] Yes
- [x] No


CharlieM312:
@ahnpnl PR ready for review

ahnpnl (approved):

PR #4053, opened 2026-08-19
build(deps): remove @typescript-eslint packages

CharlieM312:
## Summary

Remove both @typescript-eslint packages 

## Test plan

Tests pass

- [ ] Yes
- [ ] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,8 +88,6 @@
     "@types/jest": "^30.0.0",
     "@types/node": "^24.13.3",
     "@types/semver": "^7.8.0",
-    "@typescript-eslint/eslint-plugin": "^8.67.0",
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -57,6 +57,14 @@
       ],
       "enabled": false
     },
+    {
+      "matchPackageNames": [
```
ahnpnl: I would prefer to remove eslint-plugin and parser from the list of deps in `package.json` for this PR because they are installed by umbrella package `typescript-eslint`

ahnpnl: And we don’t change renovate bot config at all

CharlieM312: ok so just remove eslint-plugin and parser that makes things easier

ahnpnl (approved):

PR #4118, opened 2026-09-10
build(deps): keep consistent angular version in example apps

CharlieM312:
## Summary

Bumps angular deps to match @angular/core (in v20 and v21 example apps), as well as @angular/common versions in the monorepo example. More of a nice to have but is good for consistency and won't break anything

## Test plan

Tests should pass given there is no major version upgrade happening here

- [ ] Yes
- [x] No


CharlieM312:
@ahnpnl PR ready for review

ahnpnl (approved):

PR #4124, opened 2026-09-13
fix(transformer): avoid prototype pollution in sha1 cache

kamalesh404:
Fixes prototype pollution where \cacheKey in cache\ checks the prototype chain. \__proto__\ etc. would incorrectly hit. Use \Object.create(null)\ for the cache and \hasOwnProperty\ check. No API change.

**Update (addressing review feedback):** Removed the unreachable sha1 cache entirely — \getCacheKey\ always passes two arguments to \sha1\, making \canCache\ permanently false, so the original cache was dead code. This eliminates both the dead code and the prototype-pollution-prone code path in one change. The \sha1\ function is now a pure hash with no cache state.


ahnpnl:
@codex review

